PR #5, opened 2019-07-15
Doxy

teoperisanu:


.gitignore
```diff
@@ -0,0 +1,13 @@
+CMakeLists.txt.user
+CMakeCache.txt
```
dNechita: I recommend to not ignore the CMake generated files which are at the same level as the source files. This can encourage in-source build which is not a good practice. If an in-source build happens by accident, users could easily see which files have been generated (with git status) and remove them. But this won't be possible if the files are ignored.

adisuciu:
Can you integrate this with travis so the documentation gets automatically uploaded to a website. Something like this ? https://analogdevicesinc.github.io/libiio/ 

@DanielGuramulta

DanielGuramulta:
Documentation is available at: https://analogdevicesinc.github.io/libm2k/. Travis CI will also build the documentation and check if it is correct. The build will fail if the documentation can not be built successfully (undocumented parameter, undocumented method etc). Currently the build will fail because there are a lot of warnings when the documentation is generated. When pull-request are merged into master the documentation will be updated by Travis CI.

AlexandraTrifan:
Could you rebase and fix the conflicts and then force push on the doxy branch?

PR #15, opened 2019-08-23
Digital and Analog In / Out: Return and push interleaved samples and drop std::vector

AlexandraTrifan:
-Add methods for push and getSamples which accept raw pointers. 
This is useful in scenarios where std libraries are not available.

-Add methods which send and get interleaved data. 


tfcollins:
@AlexandraTrifan please let me know when I can test things on the MATLAB side

AlexandraTrifan:
We need to merge a branch with fixes into master. After that, this can be tested, reviewed and merged.

PR #21, opened 2019-09-14
Create M2k object based on iio_context

teoperisanu:
Signed-off-by: Teo Perisanu <Teo.Perisanu@analog.com>

### DIFF
--- a/include/libm2k/context.hpp
+++ b/include/libm2k/context.hpp
@@ -75,7 +75,7 @@ public:
 	std::string getContextAttributeValue(std::string attr);
 	std::string getContextDescription();
 	std::string getSerialNumber();
-	std::unordered_set<std::string> getAllDevices(iio_context *ctx);
+	std::unordered_set<std::string> getAllDevices();
 
 	libm2k::devices::M2k* toM2k();
 
--- a/include/libm2k/contextbuilder.hpp
+++ b/include/libm2k/contextbuilder.hpp
@@ -57,9 +57,13 @@ public:
 	*/
 	static std::vector<std::string> listDevices();
 	/**
+	 * @private
+	 */
+	static Context* deviceOpen(const char*);
+	/**
 	* @private
 	*/
-	static Context* deviceOpen(const char*);
+	static Context* deviceOpen(struct iio_context*, const char*);
 	/**
 	* @private
 	*/
@@ -67,12 +71,15 @@ public:
 	/**
 	* @private
 	*/
-	static M2k* m2kOpen(const char *);
+	static M2k* m2kOpen(const char*);
+	/**
+	* @private
+	*/
+	static M2k* m2kOpen(struct iio_context*, const char*);
 	/**
 	* @private
 	*/
 	static M2k* m2kOpen();
-
 	/**
 	* @private
 	*/
@@ -99,6 +106,7 @@ private:
  * @{
  */
 
+
 /**
 * @private
 */
@@ -106,6 +114,7 @@ LIBM2K_API Context* deviceOpen()
 {
 	return ContextBuilder::deviceOpen();
 }
+
 
 /**
  * @private
@@ -117,6 +126,15 @@ LIBM2K_API Context* deviceOpen(const char* uri)
 
 
 /**
+ * @private
+ */
+LIBM2K_API Context* deviceOpen(struct iio_context* ctx, const char* uri)
+{
+	return ContextBuilder::deviceOpen(ctx, uri);
+}
+
+
+/**
  * @brief Open a device based on its uri
  * @param uri Describe the location of the context
  * @return M2k object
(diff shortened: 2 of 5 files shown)


include/libm2k/contextbuilder.hpp
```diff
@@ -99,6 +110,7 @@ class LIBM2K_API ContextBuilder {
  * @{
  */
 
+
```
commodo: nitpick: extra line

include/libm2k/contextbuilder.hpp
```diff
@@ -107,6 +119,7 @@ LIBM2K_API Context* deviceOpen()
 	return ContextBuilder::deviceOpen();
 }
 
+
```
commodo: nitpick: extra line

include/libm2k/contextbuilder.hpp
```diff
@@ -116,6 +129,15 @@ LIBM2K_API Context* deviceOpen(const char* uri)
...
 
+/**
+ * @private
+ */
+LIBM2K_API Context* deviceOpen(struct iio_context *ctx)
+{
+	return ContextBuilder::deviceOpen(ctx);
+}
+
+
```
commodo: nitpick: extra line

src/contextbuilder.cpp
```diff
@@ -84,6 +84,62 @@ std::vector<std::string> ContextBuilder::listDevices()
 	return uris;
 }
 
+std::string ContextBuilder::getUri(struct iio_context* ctx)
+{
+	if(ctx == nullptr)
```
commodo: it's a good idea to be consistent with style;
for example: here you are using 
```
if()
{
....
}
```

while in other places, it's 
```
if () {
}
```

whatever the style, it is good to use the one that was used initially when the file was written, and also to keep the same style in a change

src/contextbuilder.cpp
```diff
@@ -130,6 +201,20 @@ Context* ContextBuilder::deviceOpen()
 	return deviceOpen(lst.at(0).c_str());
 }
 
+M2k *ContextBuilder::m2kOpen(struct iio_context* ctx)
+{
+	auto dev = deviceOpen(ctx);
+	if (!dev) {
```
commodo: style-wise, some prefer to not add brackets when the block is a single line;
this is for if() or for(), while() blocks;

so, you could do
``` 
if (!dev)
    return nullptr;
```

the brackets usually add a new line, which is not usually needed

src/contextbuilder.cpp
```diff
@@ -84,6 +84,62 @@ std::vector<std::string> ContextBuilder::listDevices()
...
+std::string ContextBuilder::getUri(struct iio_context* ctx)
+{
+	if(ctx == nullptr)
+        {
+		throw_exception(EXC_INVALID_PARAMETER, "Invalid iio context");
+        }
+	std::string uri;
+	std::string context_type(iio_context_get_name(ctx));
+
+	if(context_type == "network")
```
commodo: for a getter, this looks like it's doing a lot;
well, a lot of stuff that libiio is doing;

maybe try to create a context object with libiio, and if that works, return the URI as string
usually, and URI is passed on from a user, so we should have the original/complete URI already available somewhere

commodo:
the commit could use a description: something explaining why this change is needed, what it does, what it solves; a general overview;
sometimes changes in git history are harder to understand after 1-2 years why they were added;

mihaiolteanu:
@commodo , please review commit [51ff666](https://github.com/analogdevicesinc/libm2k/pull/21/commits/51ff6668479dbfa50de5fae9b3c039e6c8fa14e4), also.

include/libm2k/context.hpp
```diff
@@ -75,7 +75,7 @@ class LIBM2K_API Context {
 	std::string getContextAttributeValue(std::string attr);
```
commodo: this commit is missing a `Signed-off-by` tag;
do we need/want to use this for this project? different projects have different rules and legal-stuff
if yes, then it may be an idea to add it;

mhennerich: We should follow the sign-off process on all our repositories.
With this tag you certify the Developer's Certificate of Origin.

Please take a look here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=HEAD#n418

src/contextbuilder.cpp
teoperisanu: Now, with commit 5f6c11f, the style is consistent.

src/contextbuilder.cpp
teoperisanu: I consider using the brackets in a single line block is maintainability-wise, because adding more lines afterwards, without adding curly braces could cause a lot of logical errors.

src/contextbuilder.cpp
teoperisanu: You are right. The problem was fixed in commit 5f6c11f.

include/libm2k/contextbuilder.hpp
teoperisanu: The line was added in order to be style-consistent.

include/libm2k/contextbuilder.hpp
teoperisanu: The line was added in order to be style-consistent.

include/libm2k/contextbuilder.hpp
teoperisanu: The lines were added in order to be style-consistent.

commodo:
generally looks good;
@mihaiolteanu - just one comment added; please add the `Signed-off-by` tag;
@teoperisanu - style changes are fine; usually they are recommended to be done in separate commits; reason is: after a while, commits that add functionality and style-changes are noisy and hard to look through (when looking at the git history); this is something that will become more obvious over time; for this commit, it is fine as-is

include/libm2k/context.hpp
mihaiolteanu: Done.

Issue #24, opened 2019-10-01
AnalogOut: pushRaw should not process samples


adisuciu:
pushRaw method should send data directly to the DAC without any alterations. The data is scaled by -16 before being sent to the DAC.

https://github.com/analogdevicesinc/libm2k/blob/master/src/analog/private/m2kanalogout_impl.cpp#L438

https://github.com/analogdevicesinc/libm2k/blob/master/src/analog/private/m2kanalogout_impl.cpp#L267

AlexandraTrifan:
Should be fixed with commit 1290bfbba6d381f84bedf33b96c9043f4a157f70

PR #25, opened 2019-10-01
General: Add multiple fixes.

AlexandraTrifan:
This PR introduces 3 fixes throughout the library:

- For the M2K Analog Out : The raw samples should not be altered at all before being pushed to the device. They should be pushed in the same state they were provided (only applies to raw values & pushRaw() methods) ( Related to issue #24 )
- M2K Digital : Delete the trigger object.
- M2K Hardware Trigger: Change the return type for the analog trigger level from int to double.

### DIFF
--- a/src/analog/private/m2kanalogout_impl.cpp
+++ b/src/analog/private/m2kanalogout_impl.cpp
@@ -264,7 +264,7 @@ public:
 		m_samplerate.at(1) = getSampleRate(1);
 
 		for (unsigned int i = 0; i < nb_samples; i++) {
-			raw_data_buffer.push_back(processSample(data[i], chnIdx, true));
+			raw_data_buffer.push_back(data[i]);
 		}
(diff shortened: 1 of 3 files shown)


src/analog/private/m2kanalogout_impl.cpp
```diff
@@ -264,7 +264,7 @@ class M2kAnalogOut::M2kAnalogOutImpl : public DeviceGeneric {
 		m_samplerate.at(1) = getSampleRate(1);
 
 		for (unsigned int i = 0; i < nb_samples; i++) {
```
commodo: unrelated to this PR; and looking at this in a more general-mode, is there a way to do these without doing the for-loop?
this looks very C-ish, and looks like it could benefit from more C++ constructs;

would something like this work?
```
std::vector<short> raw_data_buffer(data, data + nb_samples);
```

AlexandraTrifan: This makes sense in some cases. Let me add some changes and check more thoroughly for these in the future.

PR #32, opened 2019-10-04
M2kAnalogIn & M2kAnalogOut: get available samplerates

AlexandraTrifan:
Expose methods to get the available sample rates for the M2k ADC and the M2k DAC.

### DIFF
--- a/src/analog/private/m2kanalogout_impl.cpp
+++ b/src/analog/private/m2kanalogout_impl.cpp
@@ -100,18 +100,12 @@ public:
 
 	double getCalibscale(unsigned int index)
 	{
-		if (index >= m_dac_devices.size()) {
-			throw invalid_parameter_exception("M2kAnalogOut: No such channel available for calibscale");
-		}
-		return m_dac_devices.at(index)->getDoubleValue("calibscale");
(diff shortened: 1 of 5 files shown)


src/analog/private/m2kanalogout_impl.cpp
```diff
@@ -449,6 +449,11 @@ class M2kAnalogOut::M2kAnalogOutImpl : public DeviceGeneric {
 		return m_dac_devices.at(chnIdx)->isChannelEnabled(0);
 	}
 
+	std::vector<double> getAvailableSampleRates(unsigned int chnIdx)
+	{
+		return m_dac_devices.at(chnIdx)->getAvailableSampleRates();
```
commodo: curios: what happens at `m_dac_devices.at(chnIdx)->getAvailableSampleRates()`  when `m_dac_devices.at(chnIdx)` has no element?

commodo:
otherwise looks good

src/analog/private/m2kanalogout_impl.cpp
AlexandraTrifan: That's a good point. I added a getter that checks whether the index is valid or not.

PR #35, opened 2019-10-04
Examples & doc: Add checks to handle null objects when calling m2kOpen()

AlexandraTrifan:
The methods that handle the creation of an ADALM2000 context do not throw an exception, they just return a nullptr. This PR has some additions for the Python and C++ examples and also for the documentation.

- All the examples now check if the opened context is valid. If not, the example prints an appropriate message for the user and exits the program.
- The documentation now states that those specific methods may return a nullptr.


examples/analog/main.cpp
```diff
@@ -14,6 +14,11 @@ using namespace libm2k::devices;
 int main(int argc, char* argv[])
 {
 	M2k *ctx = m2kOpen();
+	if (!ctx) {
+		std::cout << "Connection Error: No ADALM2000 device available/connected to your PC." << std::endl;
+		return 0;
```
commodo: `return 0` means success;
non-zero means failure;

it does not make much difference what return code you return; people usually check against zero; so `return 1` may be sufficient

commodo: also, there doesn't seem to be any `return 0` at the end of this function;
curios: does C++ implicitly `return 0`?

examples/digital/main.cpp
```diff
@@ -15,6 +15,11 @@ int main()
 
 {
 	M2k *ctx = m2kOpen();
+	if (!ctx) {
```
commodo: nitpick, and strictly speaking about style;
i still don't know what general style is preferred for this project;
specifically here: is it  `if()` or `if ()` [i.e. `if__space__()`] ?

bindings/python/examples/analog.py
```diff
@@ -4,6 +4,10 @@
 import numpy as np
 
 ctx=libm2k.m2kOpen()
+if ctx is None:
+	print("Connection Error: No ADALM2000 device available/connected to your PC.")
+	exit(0)
```
commodo: i'd also maybe do `exit(1)` here

commodo:
rest looks good

examples/digital/main.cpp
AlexandraTrifan: The style is: "if () {"

examples/analog/main.cpp
AlexandraTrifan: As far as I know if the control reaches the end of the main function, then it will implicitly return 0 .

bindings/python/examples/analog.py
AlexandraTrifan: I will add the changes today

PR #37, opened 2019-10-07
ContextBuilder: define API functions as inline in order to avoid multiple definition error.

teoperisanu:
In order to avoid multiple definition error, move ContextBuilder API implementation to source file.

Signed-off-by: Teo Perisanu <Teo.Perisanu@analog.com>

### DIFF
--- a/src/contextbuilder.cpp
+++ b/src/contextbuilder.cpp
@@ -222,3 +222,48 @@ DeviceTypes ContextBuilder::identifyDevice(iio_context *ctx)
 	}
 	return type;
 }
+
+Context *libm2k::devices::deviceOpen()
+{
+	return ContextBuilder::deviceOpen();
+}
+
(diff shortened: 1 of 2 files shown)


src/contextbuilder.cpp
```diff
@@ -222,3 +222,48 @@ DeviceTypes ContextBuilder::identifyDevice(iio_context *ctx)
...
+
+void libm2k::devices::deviceClose(Context *ctx, bool deinit)
+{
+	ContextBuilder::deviceClose(ctx, deinit);
+}
+
+void libm2k::devices::deviceCloseAll()
+{
+	ContextBuilder::deviceCloseAll();
+}
```
dNechita: new empty line at the end of file is needed

teoperisanu: Line added in commit 9fd0826.

PR #39, opened 2019-10-07
Enforce coding style

DanielGuramulta:
Enforce coding style on commits.

To prevent files mixing different coding styles
```cpp
// example
if () {
    ...
}

if() {
    ...
}
```
we make Travis check the style of the files, being edited(patched) in the commit(commit range), and if something is not formatted according to the `.clang-format` files specifications, we fail the build.

For this to go through, we need:

1. to discuss the coding style and code it into `.clang-format`
2. format the existing code 
3. remove the test commit a3bc159


.clang-format
```diff
@@ -0,0 +1,112 @@
...
+PenaltyBreakAssignment: 2
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Right
+ReflowComments:  true
+SortIncludes:    true
```
dNechita: This might be annoying as if sorts the included headers alphabetically. You usually want more flexibility on order of the included files.

DanielGuramulta: We can also specify include categories, which will provide some general rules on how the included headers will be sorted.
```
example:
IncludeCategories:
  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
    Priority:        2
  - Regex:           '^(<|"(gtest|gmock|isl|json)/)'
    Priority:        3
  - Regex:           '.*'
    Priority:        1
```

.travis.yml
```diff
@@ -19,6 +19,11 @@ matrix:
       env:
         - OS_TYPE=doxygen
       addons: { apt: { packages: ["cmake", "graphviz"] } }
+    - os: linux
+      dist: xenial
```
commodo: if you don't need to force `xenial`, drop it;
at this point in time, this should be the default docker image

.travis.yml
```diff
@@ -19,6 +19,11 @@ matrix:
...
+      env:
+        - OS_TYPE=clangformat
+      addons: { apt: { packages: ["clang-format-6.0"], sources: ["llvm-toolchain-trusty-6.0"] } }
```
commodo: `llvm-toolchain-trusty-6.0` ?
curios: is this correct?
it's a xenial image

CI/travis/clangformat.sh
```diff
@@ -0,0 +1,26 @@
...
+then
+COMMIT_RANGE=HEAD~1
+fi
+
+git diff --name-only --diff-filter=d $COMMIT_RANGE | while read -r file; do
+        /usr/bin/clang-format-6.0 -i "$file"
+done;
+
+git diff --exit-code || {
+echo_red "The code is not properly formatted!"
```
commodo: nitpick: indent this block?

CI/travis/clangformat.sh
```diff
@@ -0,0 +1,26 @@
...
+
+############################################################################
+# Check if the files modified in the current commit / commit range respect
+# the coding style defined in the .clang-format file
+############################################################################
+COMMIT_RANGE=$TRAVIS_COMMIT_RANGE
+
+if [ -z "$TRAVIS_PULL_REQUEST_SHA" ]
+then
+COMMIT_RANGE=HEAD~1
```
commodo: nitpick: indent this block?

commodo:
Some commits could use some descriptions

.travis.yml
DanielGuramulta: `dist:xenial` was added by mistake, in another place I used the same addons but the `dist` was forced to `trusty` for the whole matrix. Fixed the issue.

CI/travis/clangformat.sh
DanielGuramulta: Fixed the indentation.

CI/travis/clangformat.sh
DanielGuramulta: Fixed the indentation.

.travis.yml
```diff
@@ -19,6 +19,10 @@ matrix:
       env:
         - OS_TYPE=doxygen
       addons: { apt: { packages: ["cmake", "graphviz"] } }
+    - os: linux
+      env:
+        - OS_TYPE=clangformat
+      addons: { apt: { packages: ["clang-format-6.0"], sources: ["llvm-toolchain-xenial-6.0"] } }
```
commodo: curios: would this work?  `addons: { apt: { packages: ["clang-format-6.0"], sources: ["llvm-toolchain-${TRAVIS_DIST}-6.0"] } }`

also, is ther a generic `llvm-toolchain-6.0` addon?
i would not add `xenial` if possible; otherwise, it's probably better to add `dist: xenial` back to make sure things stay consistent (sorry about that, if it happens)

commodo:
commits could use a description

PR #40, opened 2019-10-08
M2kCalibration: Fix the calibration procedure, which was broken in PR #25

AlexandraTrifan:
PR #25 removed the raw value processing from the AnalogOut methods.
One of those methods was used in the calibration procedure. Only for this procedure, the specific processing is needed in order to have the board functional.

### DIFF
--- a/src/private/m2kcalibration_impl.cpp
+++ b/src/private/m2kcalibration_impl.cpp
@@ -473,7 +473,8 @@ out_cleanup:
 		m_ad5625_dev->setDoubleValue(1, 2048, "raw", true);
 
 		// write to DAC
-		std::vector<short> vec_data(256, 0);
+		int16_t value = processRawSample(0);
+		std::vector<short> vec_data(256, value);
 		std::vector<std::vector<short>> vec_data_all;
 		__try {
 			vec_data_all.push_back(vec_data);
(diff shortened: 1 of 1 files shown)


src/private/m2kcalibration_impl.cpp
```diff
@@ -776,6 +778,13 @@ class M2kCalibration::M2kCalibrationImpl {
 		return true;
 	}
 
+	int16_t processRawSample(int16_t value)
+	{
+		// This should go away once channel type gets
+		// changed from 'le:S16/16>>0' to 'le:S12/16>>4'
+		return ((-value) << 4);
```
commodo: is this a driver change? or a change that should be done in the driver?
if yes, maybe we can do it now and use that as base

AlexandraTrifan: I don't have a lot of details about this change. We can investigate it further.

commodo: @mhennerich 
any thoughts/input here?

mhennerich: All our DAC HDL cores expect the realbit MSB aligned with the MSB of the storage bit.
Even it the change the type, which is basically just a string, I don't think anything would go away here...

commodo: then should we remove the string as it may be misleading?
or change it to something else?

mhennerich: Ack - change comment to:
12-bit DAC values must be 16-bit MSB aligned.

AlexandraTrifan: Comment updated

PR #43, opened 2019-10-15
M2kAnalogOut: Fix cyclic mode for M2kAnalogOut.

AlexandraTrifan:
This fixes a mistake added in one of the previous commits. The cyclic mode attribute was no longer taken into consideration because of the wrong getter.


commodo:
Just one nitpick: the commit description could be s bit more descriptive about why/what is being fixed (or broken).

It’s obvious from the code.
But still could be useful.
Fine to also leave it like this.
No strong preference.

Issue #36, opened 2019-10-07
deviceOpen-ing a device not in m_dev_map throws 'std::logic_error'


mihaiolteanu:
I have a valid network device (LiDAR), that can be opened and handled via libiio,

```
iio_context* ctx = iio_create_context_from_uri("ip:10.48.65.117");
cout << iio_context_get_name(ctx);

=> network
```

Trying to do the same with libm2k throws a 'std::logic_error',
```
Context *context = ContextBuilder::deviceOpen("ip:10.48.65.117");

=> terminate called after throwing an instance of 'std::logic_error'
         what():  basic_string::_M_construct null not valid

Compilation aborted (core dumped) at Mon Oct  7 11:55:08
```

Expected result: a generic device (product, actually) is returned which would support at least some identification functions from iiolib (list of devices, context name, context attributes, etc..)

Short code-jumping without actually knowing where the problem comes from,

- `identifyDevice` returns `Other` if device is not in `m_dev_map` list, which is the case of the device from above (ok).

- `m_dev_name_map.at(type)` in `buildDevice` should then return `Generic` (ok).

- `ContextImpl` constructor then receives the device `uri` together with the `Generic` name for this device. I'm not sure what the `Generic` name is used for as it doesn't appear to be used in this constructor. So the fact that this device is not in the list of known devices is not taken into consideration?! Maybe you're assuming this is a known device and try to stuff it doesn't actually support?! I'm just assuming here.

AlexandraTrifan:
Fixes are merged to master.

PR #51, opened 2019-10-28
Context: General fixes and additions

AlexandraTrifan:
1. Add getters for Analog In device count, Analog Out device count, DMM device count.
2. Some IIO devices don't have a name, so we check the id instead. See bug #36 .
3. Add a Context method that allow the user to check the library version.


src/private/context_impl.cpp
```diff
@@ -285,7 +285,13 @@ class Context::ContextImpl {
 		std::unordered_set<std::string> device_list;
 		for (unsigned int i = 0; i < nb_devices; ++i) {
 			auto dev = iio_context_get_device(m_context, i);
-			device_list.emplace(std::string(iio_device_get_name(dev)));
```
commodo: related to this change: why not do `return Utils::getAllDevices(m_context)` or something like that?

does not have to be in this PR; but it looks doable; unless my coffee did not kick-in yet and i missed something :)

commodo:
rest looks good

src/private/context_impl.cpp
AlexandraTrifan: Makes sense.
I added the changes.

PR #54, opened 2019-11-08
M2k Trigger: Add support for firmware 0.24 trigger features.

AlexandraTrifan:
This PR introduces support for the new firmware version and some other necessary utilities.

- The v0.24 trigger structure is different than all the other FW version so we need to keep the library compatible with both old and new versions. 
- Add some fixes on the trigger and general side (related to attribute data type - oversampling ratio, trigger hysteresis)
- Move the trigger implementation into the appropriate namespace. ("libm2k", not "libm2k::analog", not "libm2k::digital")


src/utils/utils.cpp
```diff
@@ -255,3 +266,29 @@ DEVICE_DIRECTION Utils::getIioDeviceDirection(struct iio_device* dev)
...
+	} else if (v2.size() < v1.size()) {
+		std::string surplus("0", (v1.size() - v2.size()));
+		v2.append(surplus);
+	}
+	if (v1 == v2) {
+		return 0;
+	} else if (v1 < v2) {
+		return 1;
+	}
+	return -1;
```
dNechita: A more compact for of this can be this one liner:
`return v1.compare(v2);`

include/libm2k/analog/m2khardwaretrigger.hpp
```diff
@@ -49,7 +49,7 @@ class LIBM2K_API M2kHardwareTrigger
 	/**
 	 * @private
 	 */
-	M2kHardwareTrigger(struct iio_context *ctx);
+	M2kHardwareTrigger(struct iio_context *ctx, std::string fw_version);
```
dNechita: fw_version can be const ref

src/analog/private/m2kanalogin_impl.cpp
```diff
@@ -42,14 +42,14 @@ using namespace std::placeholders;
...
-	M2kAnalogInImpl(iio_context * ctx, std::string adc_dev, bool sync) :
+	M2kAnalogInImpl(iio_context * ctx, std::string adc_dev, bool sync, M2kHardwareTrigger *trigger) :
 		DeviceIn(ctx, adc_dev),
 		m_need_processing(false),
 		m_trigger(nullptr)
 	{
 		m_m2k_fabric = make_shared<DeviceGeneric>(ctx, "m2k-fabric");
 		m_ad5625_dev = std::make_shared<DeviceGeneric>(ctx, "ad5625");
-		m_trigger = new M2kHardwareTrigger(ctx);
+		m_trigger =  trigger;
```
dNechita: you're setting the m_trigger twice. First was in the initializer list.
What happens if someone passes nullptr for trigger? M2kAnalogInImpl will work with limited functionality? or crash? should this constructor throw?

AlexandraTrifan: You're right, something is missing here.
How I see things: one way to go would be to throw an exception here: https://github.com/analogdevicesinc/libm2k/pull/54/files#diff-b7d2236a13d61e896bfdcd1529c58fddR66 . An M2K should always have a valid trigger structure, if the object is not correctly created then the M2K Context should not work at all (so we throw the exception right at the beginning).

Any thoughts on this? @dNechita @adisuciu ?

src/utils/utils.cpp
AlexandraTrifan: Updated.

src/analog/private/m2kanalogin_impl.cpp
AlexandraTrifan: Changed to throw an exception where it first creates the trigger object .

Issue #31, opened 2019-10-03
libiio error codes are not handled


mihaiolteanu:
libiio provides error codes for most of it's calls, but these errors are mostly not taken into consideration in libm2k. That is, the return value of the iio call is either not used anywhere, or the error code and/or string is not logged nor thrown in the exception.

Some examples,
```
/* In devicegeneric_impl.cpp. Return value not considered.
if (Context::iioDevHasAttribute(m_dev, attr)) {
    iio_device_attr_read(m_dev, attr.c_str(), value, sizeof(value));  /* <------ */
} else {
    throw_exception(EXC_INVALID_PARAMETER, dev_name +   " has no " + attr + " attribute");
```
```
/* In devicegeneric_impl.cpp. Return value considered, but not handled. */
/* The thrown string does not contain the reason for the error. */
int ret = iio_context_get_attr(m_context, attrIdx, &name, &value);
    if (ret < 0) {
        throw_exception(EXC_RUNTIME_ERROR, "Device: Can't get context attribute " +
```
```
/* In context_impl.cpp. Return value not considered. */
iio_context_get_version(m_context, &ctx_major, &ctx_minor, ctx_git_tag);
```
etc, etc.

You can get the error string from the error code with the `iio_strerror` function provided by libiio. The return value for the functions that do return an error code is 

>  * @return On success, 0 is returned
>  * @return On error, a negative errno code is returned

There are currently 134 defined errors in libiio.

This is just ignoring useful information that could prove useful in case of bugs and/or abnormal behavior from the library.

PR #55, opened 2019-11-08
WIP: General: Rework the exceptions across the entire library.

AlexandraTrifan:
- This commit adds error handling for the libiio calls. Previously, the
error codes from IIO were not handled. As they provide additional info,
we now add the error string to the already existing exception string (see #31).
- Also, the file name and line number are also included in the
exception string.

### DIFF
--- a/include/libm2k/m2kexceptions.hpp
+++ b/include/libm2k/m2kexceptions.hpp
@@ -79,8 +79,11 @@ public:
 
 static std::exception e;
 
-static void throw_exception(libm2k::M2K_EXCEPTION exc_type, std::string exception)
+static void throw_exception(libm2k::M2K_EXCEPTION exc_type, std::string exception,
+			    std::string file, int line)
 {
+	exception += "; File: " + file;
+	exception += "; Line: " + std::to_string(line);
 #if _EXCEPTIONS || defined(__cpp_exceptions)
 	switch (exc_type) {
 	case libm2k::EXC_OUT_OF_RANGE: {
--- a/src/analog/private/dmm_impl.cpp
+++ b/src/analog/private/dmm_impl.cpp
@@ -89,7 +89,7 @@ public:
 			value = channel->getDoubleValue("input");
 		} else {
 			throw_exception(EXC_INVALID_PARAMETER, "DMM: Cannot read channel " +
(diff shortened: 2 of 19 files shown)


mihaiolteanu:
You think it would be better to maybe have the `strerror` conversion right inside the `throw_exception` function?
```
static void throw_exception (libm2k::M2K_EXCEPTION exc_type, 
                            std::string exception,
                            int err_number,         // <----- add this new parameter
                            std::string file, int line) {
    exception += "std::strerror(err_number);"       <---- handle the `strerror` call here
    exception += "; File: " + file;
    exception += "; Line: " + std::to_string(line);
    ....
}
```

This would simplify all the code snippets that look like this,
```
if (ret < 0) {
    errstr = "Error: get version ";
    // errstr += std::strerror(-ret);   <---- remove this line and use -ret in the `throw_exception` call
    goto out_cleanup;
}
```

But most importantly it would also handle cases where `iiolib` directly sets the `errno` but does not return it as a negative integer. And that means you could also throw the error string in cases such as this,
```
if (channel->hasAttribute("raw")) {
    value = channel->getDoubleValue("raw");
} else if (channel->hasAttribute("processed")) {
    value = channel->getDoubleValue("processed");
} else if (channel->hasAttribute("input")) {
    value = channel->getDoubleValue("input");
} else {
    throw_exception(EXC_INVALID_PARAMETER,
        "DMM: Cannot read channel " + getName() + " : " + chn_name,
        errno,  <---- add this argument in the throw_exception call which is the actual err code set by iiolib
        __FILE__,
        __LINE__);
}
```

In short, in cases where the error code is returned directly by the iiolib call, pass that to the `throw_exception` function. Where iiolib does not return an error code but sets the `errno` directly, pass that to the `throw_exception` function. And, finally, handle the error codes and the actual error string thrown to the user in just one place, the `throw_exception` function.

AlexandraTrifan:
Makes sense. Will work on this soon.

mihaiolteanu:
I can modify it, if you want to.
Please let me know.

teoperisanu:
I think this PR is outdated. The rework of the exception mechanism across the library was updated in #168 and #173.

PR #64, opened 2019-11-19
Device In: Handle all types of data sizes.

AlexandraTrifan:
Allow libm2k to handle data acquisition on devices with data sizes different from data size on "m2k-adc" in ADALM2000.

### DIFF
--- a/src/utils/private/buffer_impl.cpp
+++ b/src/utils/private/buffer_impl.cpp
@@ -365,6 +365,11 @@ public:
 
 	const short* getSamplesRawInterleaved(int nb_samples)
 	{
+		return static_cast<const short*>(getSamplesRawInterleavedVoid(nb_samples));
+	}
+
+	void *getSamplesRawInterleavedVoid(int nb_samples)
+	{
 		bool anyChannelEnabled = false;
 		if (Utils::getIioDeviceDirection(m_dev) != INPUT) {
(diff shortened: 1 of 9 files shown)


src/utils/private/buffer_impl.cpp
```diff
@@ -364,6 +364,12 @@ class Buffer::BufferImpl {
 
 
 	const short* getSamplesRawInterleaved(int nb_samples)
+	{
+		const short* p_dat = (const short*) getSamplesRawInterleavedVoid(nb_samples);
```
commodo: why not just do `return getSamplesRawInterleavedVoid(nb_samples);` ?

at least C [I think C++ as well] doesn't complain if you allow implicit casts to/from `void *`
another mechanism, may be to use `reinterpret_cast`; i am not sure about it's added value of C-type casts, but it looks more C++-ish

commodo:
i do feel that the commit description could use a bit more info;

src/utils/private/buffer_impl.cpp
DanielGuramulta: C++ allows implicit conversions TO a void * type, but to convert FROM a void * type requires an explicit cast. In this case (const short *) if you're using c style cast, or an static_cast if you're using c++ style casts.

AlexandraTrifan:
Added the changes and some description to the commit.

commodo:
> Added the changes and some description to the commit.

thank you :)

PR #68, opened 2019-11-21
M2kDigital: Fix TX enableChannel method

AlexandraTrifan:
After recent changes to the Device In/Out/Generic channels
scanning system, the enableChannel() method no longer works without
specifying the TX device for channel number checking. That is because
the GenericDevice(m2k-logic-analyzer) only has input channels.


commodo:
many thanks for the problem description

PR #69, opened 2019-11-22
Add Lidar

mihaiolteanu:
Add Lidar to the list of boards discoverable by the ContextBuilder.

Use the DeviceGeneric and DeviceIn classes to implement the Lidar requirements from the Lidar API page [1] and expose the relevant interfaces to the user.

Initialize the board in the same way the iio-oscilloscope does with the Lidar plugin (some code is actually identical with that), but also enable Channel4 by default, which should always be enabled.

Reading the buffer returns a list of all the enabled channels together with the requested number of samples.

[1] [Lidar API](https://confluence.analog.com/pages/viewpage.action?spaceKey=SDG&title=LiDAR+API)

Signed-off-by: Mihai.Olteanu <Mihai.Olteanu@analog.com>


commodo:
maybe do a rebase on top of master first
this branch has a few merge commits that are un-needed

commodo:
a bit curios also: why Lidar in libm2k ?
i thought libm2k was generic M2k support, and not App-specific

mihaiolteanu:
> maybe do a rebase on top of master first
> this branch has a few merge commits that are un-needed

I've already merged master on lidar a few times. Rebase would make a big mess.

mihaiolteanu:
> a bit curios also: why Lidar in libm2k ?
> i thought libm2k was generic M2k support, and not App-specific

Now it is generic support for everything. Repurposed for a brighter future.
I don't have any more info than that.

src/private/lidar_impl.cpp
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        ~LidarImpl() {
+                ;
+        }
+    
+private:
+        struct DeviceGeneric* pulse_dev;
+        struct DeviceGeneric* afe_dev;
+        struct DeviceIn* adc_dev;
```
AlexandraTrifan: Do we need "struct" here?

mihaiolteanu: Good catch! 
Removed with the latest commit.

include/libm2k/lidar.hpp
```diff
@@ -27,141 +25,54 @@ namespace libm2k {
...
             };
             
-            // All APD channels.
             enum APDChannel {
                 VOLTAGE0,
                 VOLTAGE1
             };
 
             // **************** Data aquisition ****************
-
```
commodo: you can squash this commit into the first one

commodo:
Commits are missing `Signed-off-by` tags.

commodo:
so; after going through all the commits, i would say to squash all of them into one;
the series is hard to follow and holds no value;
it kind of shows that there were many changing of minds/thoughts during development (which is fine);

however, we don't need to commit them into the `master` history
which is why i think that maybe just squashing them into a single commit would be fine;

mihaiolteanu:
@commodo, all remarks implemented.

include/libm2k/contextbuilder.hpp
```diff
@@ -30,6 +30,7 @@
 enum ContextTypes {
 	CtxFMCOMMS,
 	CtxM2K,
+        CtxLIDAR,
```
commodo: this should be indented with tabs;
it's indented with spaces;

when adding new code, it's good to preserve the original styling and spacing

mihaiolteanu: Done

include/libm2k/lidar.hpp
```diff
@@ -0,0 +1,52 @@
...
+			void enableChannelSequencer();
+			void disableChannelSequencer();
+			void setSequencerPulseDelay(long long ns);
+
+			void setChannelSequencerOpModeManual();
+			void setChannelSequencerOpModeAuto();
+			void setChannelSequencerOrderAutoMode(const char* order);
+			void setChannelSequencerOrderManualMode(const char* order);
+    
+			// void enableTempCompensation();
```
commodo: these methods are commented out;
if they won't be implemented ever, maybe let's remove them from here;

if they will be implemented later, it looks good if you add them when they will be implemented;

either way; i don't have a strong opinion about leaving this as-is

src/private/lidar_impl.cpp
```diff
@@ -0,0 +1,191 @@
...
+using namespace libm2k::utils;
+
+class Lidar::LidarImpl : public ContextImpl {
+public:
+	LidarImpl(std::string uri, iio_context* ctx, std::string name, bool sync) :
+		ContextImpl(uri, ctx, name, sync) {
+
+		pulse_dev = new DeviceGeneric(ctx, "7c700000.axi-pulse-capture");
+		afe_dev	  = new DeviceGeneric(ctx, "ad5627");
+		adc_dev	  = new DeviceIn     (ctx, "axi-ad9094-hpc");
```
commodo: these values seem hard-coded;
for now, maybe this is fine as is, but maybe later a better dynamic mechanism would be better;
if you change the address from `7c700000` to something else, it no longer works;

src/private/lidar_impl.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+		setAllIIoAttributesToDefaultValues();
+	}
+
+	void initialize() {
+		;
+	}
+    
+	void channelEnableDisable(string channel, bool enable) {
+		if	(channel == "voltage0")
```
commodo: maybe here a for-loop would look nicer;
something like (pseudo-code):
```
for (i = 0; i < MAX_CHANELS; i++) {
     if (channel == "voltage" + i) {
        adc_dev->enableChannel(i, enable, false);
        return;
    }
```

src/private/lidar_impl.cpp
```diff
@@ -0,0 +1,191 @@
...
+		else if (channel == "voltage1")
+			adc_dev->enableChannel(1, enable, false);
+		else if (channel == "voltage2")
+			adc_dev->enableChannel(2, enable, false);
+		else if (channel == "voltage3")
+			adc_dev->enableChannel(3, enable, false);
+		else if (channel == "voltage4")
+			adc_dev->enableChannel(4, enable, false);
+		else
+			return;		    // invalid channel
```
commodo: should this signal an error?
or through an exception?

src/private/lidar_impl.cpp
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	vector<string> enabledChannelsList() {
+		vector<string> enabled_channels;
+		if (adc_dev->isChannelEnabled(0, false))
```
commodo: maybe do for loop here as well?
```
```
for (i = 0; i < MAX_CHANELS; i++) {
     if (if (adc_dev->isChannelEnabled(i, false)) {
       enabled_channels.push_back("voltage" + i); 
    }
```

**Comment by commodo** [comment]
Github forced me to use the `review` mechanism; which is why all comments have come up suddenly

File: include/libm2k/lidar.hpp (hunk 4 above)
Comment (mihaiolteanu) [reply include/libm2k/lidar.hpp]: Comments don't hurt anyone. I will leave them as such.

**Comment by commodo** [comment]
rest looks good

File: include/libm2k/lidar.hpp (hunk 4 above)
Comment (commodo) [reply include/libm2k/lidar.hpp]: agreed;
they are usually more about etiquette

File: src/private/lidar_impl.cpp (hunk 5 above)
Comment (mihaiolteanu) [reply src/private/lidar_impl.cpp]: I'm not sure what you mean here. 
These are the devices available on the board. What else would one use instead of "hard-coded values"?

File: src/private/lidar_impl.cpp (hunk 6 above)
Comment (mihaiolteanu) [reply src/private/lidar_impl.cpp]: Sure. The for loop looks shorter, but `"voltage" + i` looks like a hack. I will leave it like this.

File: src/private/lidar_impl.cpp (hunk 7 above)
Comment (mihaiolteanu) [reply src/private/lidar_impl.cpp]: Nothing happens if you try to enable a channel that does not exist. Looks clean to me.

File: src/private/lidar_impl.cpp (hunk 8 above)
Comment (mihaiolteanu) [reply src/private/lidar_impl.cpp]: Same as above.

File: src/private/lidar_impl.cpp (hunk 7 above)
Comment (AlexandraTrifan) [reply src/private/lidar_impl.cpp]: Should we let the user know he didn't do something right? 
My opinion is that we could throw here something like: throw_exception(EXC_INVALID_PARAMETER, "Lidar: Can't find channel" + channel);
Might be useful? 
For example: If the user has a typo in the channel name, he'll think that he enabled one channel, when in fact, nothing happened.

File: src/private/lidar_impl.cpp (hunk 5 above)
Comment (AlexandraTrifan) [reply src/private/lidar_impl.cpp]: We can't really determine dynamically which are the IIO devices we need, other than specifying the exact name. This looks ok to me for now. If we find some other mechanism we can do the changes later on.

It's also done here: https://github.com/analogdevicesinc/libm2k/blob/master/src/analog/private/m2kanalogin_impl.cpp#L50 where we need those 2 devices for configuration.

Comment (commodo) [reply src/private/lidar_impl.cpp]: > I'm not sure what you mean here.

so, the number `7c700000` comes from here https://github.com/analogdevicesinc/linux/blob/master/arch/arm/boot/dts/zynq-zc706-adv7511-fmclidar1.dts#L59

which means, that the `axi-pulse-capture` IP is mapped at that specific address;

typically, customers take our designs as reference, and change stuff; sometimes it's it in the kernel, or device-tree, or similar;
sometimes, they also change addresses;
when they stuck, they ask questions here:  https://ez.analog.com/   [and we have to answer stuff there]

i'm seeing quite a potential for some customer-questions regarding that address;

Comment (commodo) [reply src/private/lidar_impl.cpp]: > We can't really determine dynamically which are the IIO devices we need, other than specifying the exact name. This looks ok to me for now. If we find some other mechanism we can do the changes later on.

Fine by me: I am raising some points here about the relationship between Linux + DT & this code.
But we do need to keep in mind this, when customers ask us stuff.

Another way to do this, is via some config mechanism.
But, that is outside of the scope of this changeset.

Comment (mihaiolteanu) [reply src/private/lidar_impl.cpp]: So, regarding this device only that address is subject to change? The device name is not subject to change, right?

If that's the case, then sure, we can do a search for the device name only, regardless of the address. Maybe it's reasonable to implement that inside the libm2k itself and not on Lidar? Meaning, you can do a `DeviceGeneric(ctx, "axi-pulse-capture")` instead and drop the address part from the object creation.

File: src/private/lidar_impl.cpp (hunk 6 above)
Comment (commodo) [reply src/private/lidar_impl.cpp]: > Sure. The for loop looks shorter, but `"voltage" + i` looks like a hack. I will leave it like this.

this can be discussable [whether it's a hack or not] ; the current implementation relies on the fact that the number of channels will always be 5 (0 to 4);

1. assuming that some newer design will use fewer channels -> this may cause weird/undetermined behavior if a customer tries to access channel "voltage4", when the max channel is `voltage1`
2. assuming that some newer design will need more channels -> this code needs to add more `if () else () ` branches;

with the approach I'm proposing, you just change the number of `MAX_CHANELS` in a single place, and make sure that the code can adapt to this configuration

File: src/private/lidar_impl.cpp (hunk 5 above)
Comment (commodo) [reply src/private/lidar_impl.cpp]: > So, regarding this device, only that address is subject to change? The device name is not subject to change, right?

We can choose to make the device-name constant.
But the address is something that should always be treated as variable/floating,

File: src/private/lidar_impl.cpp (hunk 7 above)
Comment (mihaiolteanu) [reply src/private/lidar_impl.cpp]: Depends on what the general approach is in case the error is on the user side, meaning, it's not using the library as it should be used (as I've happened to do countless times). Specifying a wrong channel, even though the user has access to all the available channels, is a user error, not a libm2k error.

If you insist on implementing with throw, I can do that.

File: src/private/lidar_impl.cpp (hunk 6 above)
Comment (mihaiolteanu) [reply src/private/lidar_impl.cpp]: Sure. Since we're discussing "what if's" in the not so distant future, a future version might use "altvoltageX' instead of "voltageX".

My implementation looks ugly as hell, the for loop is a lot cleaner but also...'smarter'. I see this as a non-issue for now.

If you insist, I can implement it either way, no problem for me. @AlexandraTrifan, @adisuciu?

File: src/private/lidar_impl.cpp (hunk 7 above)
Comment (commodo) [reply src/private/lidar_impl.cpp]: > Depends on what the general approach is in case the error is on the user side, meaning, it's not using the library as it should be used [snip]

the question is more like: if someone were to use this library as-is: what would he/she expect to see/get if something doesn't go correctly?
1. no error?
2. a return code?
3. a C++ exception?
4. something else?

my sentiment is that this is an API-style library; and these things should be accounted for;

File: src/private/lidar_impl.cpp (hunk 5 above)
Comment (commodo) [reply src/private/lidar_impl.cpp]: I just had another thought; this can be something to consider later:
```
		afe_dev	  = new DeviceGeneric(ctx, "ad5627");
		adc_dev	  = new DeviceIn     (ctx, "axi-ad9094-hpc");
```

the `ad5627` and `ad9094` names seem specific for this version of Lidar (let's say Lidar1);
but if we want to do Lidar2 [later], the parts may be different;
sometimes, some parts go obsolete [or out-of-stock], so you end up with having the replace `ad5627` with some other similar part (e.g. ad5xxx)

maybe a more generic approach would be to try to name these devices `lidar-afe` or `lidar0-afe` in the OS, so that they can be swap-able at a later point in time;

naturally, we don't need to do this now; but this is something to consider

File: src/private/lidar_impl.cpp [hunk 9]
```diff
@@ -0,0 +1,191 @@
...
+		int8_t* start = (int8_t*)adc_dev->getSamplesRawInterleavedVoid(nb_samples);
+		int8_t* sample;
+		int8_t* end   = start + (nb_samples * enabled_channels_count);
+		std::ptrdiff_t step = 1 * enabled_channels_count; // one byte/sample for Lidar
+
+		std::vector<int8_t> samples;
+		std::map<string, vector<int8_t>> result;
+
+		// For every channel that is enabled at the time of buffer refill..
+		for (vector<string>::size_type i = 0; i != enabled_channels_count - 1; i++) {
```
commodo: what happens here if `enabled_channels_count == 0`?

src/lidar.cpp
```diff
@@ -0,0 +1,76 @@
...
+void Lidar::setSequencerPulseDelay(long long ns) {
+	m_pimpl->setSequencerPulseDelay(ns);
+}
+
+void Lidar::setChannelSequencerOpModeManual() {
+	m_pimpl->setChannelSequencerOpMode("auto");
+}
+
+void Lidar::setChannelSequencerOpModeAuto() {
+	m_pimpl->setChannelSequencerOpMode("manual");
```
commodo: this looks weird;  setChannelSequencerOpModeAuto() == `m_pimpl->setChannelSequencerOpMode("manual");` ?
is this correct?

this could be correct as-is, but i am just asking

src/lidar.cpp
```diff
@@ -0,0 +1,76 @@
...
+void Lidar::disableChannelSequencer() {
+	m_pimpl->setChannelSequencerEnableDisable(false);
+}
+
+void Lidar::setSequencerPulseDelay(long long ns) {
+	m_pimpl->setSequencerPulseDelay(ns);
+}
+
+void Lidar::setChannelSequencerOpModeManual() {
+	m_pimpl->setChannelSequencerOpMode("auto");
```
commodo: same as below

src/private/lidar_impl.cpp
mihaiolteanu: Channel4 is enabled by default at initialization, so at least one channel is always enabled. Unless...c**p, the user might play with the thing and disable all channels, just to see what happens. Good call!

Ok, two possibilities here, as we've discussed before but never reached a consensus. Return an empty result or throw an exception?

src/lidar.cpp
mihaiolteanu: Most def a mix-up. Fixed! Thanks!

src/lidar.cpp
mihaiolteanu: Same as above

src/private/lidar_impl.cpp
AlexandraTrifan: In cases like this ( or the one we discussed above) my opinion is that the library should throw an exception.
In this particular scenario the following line should throw because no channel is enabled: 
`int8_t* start = (int8_t*)adc_dev->getSamplesRawInterleavedVoid(nb_samples);`
( https://github.com/analogdevicesinc/libm2k/blob/master/src/utils/private/buffer_impl.cpp#L384 )

mihaiolteanu: @AlexandraTrifan, confirmed. Thanks! This is a non-issue, then.

src/private/lidar_impl.cpp
AlexandraTrifan: I think in this case we should throw an exception. 
We don't really use return codes right now and we want the user to know there's something wrong with the way he's using this.

mihaiolteanu: Fixed by throwing an exception.

AlexandraTrifan:
Looks ok to me. 
Will wait until the Appveyor build is done and merge it.

PR #70, opened 2019-11-27
M2kAnalogOut: data streaming

teoperisanu:
Add the possibility of setting the number of kernel buffer. Multiple kernel buffers are required for streaming data. The execution time of pushing on a single channel was improved in order to achieve a continuous signal.


src/utils/private/buffer_impl.cpp
```diff
@@ -270,7 +292,11 @@ class Buffer::BufferImpl {
 			throw_exception(EXC_INVALID_PARAMETER, "Device not output buffer capable, so no buffer was created");
 		}
 
-		destroy();
+		if (nb_samples != m_last_nb_samples || cyclic) {
```
AlexandraTrifan: Can you explain this in a couple of words (some comments)? What does libiio do if we try to do this when in cyclic/non cyclic mode.

src/analog/private/m2kanalogout_impl.cpp
```diff
@@ -312,12 +349,29 @@ class M2kAnalogOut::M2kAnalogOutImpl : public DeviceGeneric {
...
+			if (isBufferEmpty) {
+				// when kernel buffers are empty both channels must be synchronized
+				setSyncedDma(true);
+			}
+		} else {
+			setSyncedDma(true);
+		}
 
 		for (unsigned int chn = 0; chn < nb_channels; chn++) {
 			std::vector<short> raw_data_buffer = {};
```
AlexandraTrifan: Can we use bufferSize on the next line instead of nb_samples/nb_channels?

teoperisanu: Modified in commit 85c9a60.

src/utils/private/buffer_impl.cpp
teoperisanu: Added some comments in commit 0377d99.

examples/analog/streaming_synchronized.cpp
```diff
@@ -0,0 +1,63 @@
...
+using namespace libm2k;
+using namespace libm2k::analog;
+using namespace libm2k::contexts;
+
+constexpr float pi = 3.14;
+constexpr double sampleRate = 75000;
+
+int main()
+{
+	M2k *context = m2kOpen("usb:5.15.5");
```
AlexandraTrifan: I think we should remove the uri from this example.

include/libm2k/analog/m2kanalogout.hpp
```diff
@@ -239,7 +248,8 @@ class LIBM2K_API M2kAnalogOut
 	* @param data A pointer to the samples
 	* @param nb_samples the number of samples
 	*
-	* @note The corresponding DAC will be powered up
+	* @note Streaming data is possible - required multiple kernel buffers
+	* @note The given channel won't be syncronyzed with the other channel
```
AlexandraTrifan: typo at synchronized

examples/analog/streaming_synchronized.cpp
teoperisanu: Removed in commit 6205d71.

include/libm2k/analog/m2kanalogout.hpp
teoperisanu: Modified in commit da68e64.

PR #71, opened 2019-11-28
General: Added LGPL v2.1 license headers

adisuciu:


AlexandraTrifan:
Looks good to me

PR #79, opened 2019-12-11
m2k-osc: Initial revision

cristina-suteu:
Signed-off-by: Cristina Suteu <cristina.suteu@analog.com>


tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,199 @@
...
+        get_samples_thread.setDaemon(True)
+        get_samples_thread.start()
+    def stop_in(self):
+        global instream
+        instream=False
+    def update_plot(self):
+        self.plot()
+        if instream is True:
+            self.after(10,self.update_plot)
+    def update_readsamples(self):
```
commodo: it's a good idea to leave some empty lines between methods/functions in Python;

tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,199 @@
...
+        global read_samples
+        read_samples.set(self.ain_samples_entry.get())
+        print(read_samples.get())
+        print(self.ain_samples_entry.get())
+
+
```
commodo: nitpick: extra line

tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,199 @@
...
+            rangelimits_ch0.set(str(ain.getRangeLimits(ain.getRange(0))))
+    def set_range_ch1(self):
+        if ain.isChannelEnabled(1) is True:
+            if ain_range_ch1_var.get()==1:
+                ain.setRange(1,1)
+            else:
+                ain.setRange(1,0)
+            rangelimits_ch1.set(str(ain.getRangeLimits(ain.getRange(1))))
+    def start_in(self):
+        global instream
```
commodo: i'm a bit fuzzy on how `global` works inside classes, but i don't know if this is a good idea in general;
because, `instream` could become available on the whole scope of this module [at least that's how it works in general]; and the some things could get mixed up [as changes happen in the code];
especially if you get more instances of the same class (e.g. `ain_tab_frame` )

maybe assign `instream` as member of the instance;
so; something like `self.instream = True`

at least then you are safe that the value of this is accessed only on this instance;

same comment for other `global` [if it applies]

commodo:
generally: a commit description would be nice, detailing the change;

tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,199 @@
+import matplotlib
```
commodo: just curios: does this work for Python3? or just Python2?
generally, it's a good idea to target Py3 as Py2 is kind-of-going away;

tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,199 @@
...
+import threading
+import time
+import ctx_tab
+import aout_tab
+LARGE_FONT = ("Verdana",12)
+SMALL_FONT = ("Vedana",10)
+
+samples_queuech0=[]
+samples_queuech1=[]
+fig=fg.Figure(figsize=(8,6))
```
commodo: it's a good idea to be consistent about style;
so, for example, when assigning stuff:   `SMALL_FONT__space__=__space__("Vedana",10)` vs `samples_queuech0__no_space__=__no_space__[]` 

not a big deal; but consistency looks nice;

tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,199 @@
...
+from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg, NavigationToolbar2Tk
+from matplotlib import style
+import threading
+import time
+import ctx_tab
+import aout_tab
+LARGE_FONT = ("Verdana",12)
+SMALL_FONT = ("Vedana",10)
+
+samples_queuech0=[]
```
commodo: [curios] generally speaking about this Python module;
is it a standalone app or a library?

depending on the case, it's a good idea to keep samples queued on the module or not;

tools/python/m2k-osc/m2k-osc.py
```diff
@@ -0,0 +1,69 @@
...
+
+def main():
+    root = tk.Tk()
+    root.title("M2K Osc")
+    root.geometry("800x850")
+    #m2k_osc_app(root).place(side="top", fill="both", expand=True)
+    m2k_osc_app(root).grid(row=0,column=0) 
+    root.mainloop()
+
+
```
commodo: nitpick:  this could do fewer lines;

tools/python/m2k-osc/m2k-osc.py
```diff
@@ -0,0 +1,69 @@
...
+
+if  __name__ == "__main__":
+    main()
+
+
```
commodo: nitpick: extra line

tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,199 @@
...
+from tkinter import ttk
+import matplotlib.figure as  fg
+import libm2k
+from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg, NavigationToolbar2Tk
+from matplotlib import style
+import threading
+import time
+import ctx_tab
+import aout_tab
+LARGE_FONT = ("Verdana",12)
```
commodo: if you remove these fonts and leave them in the app part to choose, would it work?

tools/python/m2k-osc/m2k-osc.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+import tkinter as tk
+from tkinter import ttk
+import matplotlib.figure as  fg
+import libm2k
+from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg, NavigationToolbar2Tk
+from matplotlib import style
+import threading
+import time
+from ctx_tab import *
```
commodo: it's not always a good idea to import everything;

tools/python/m2k-osc/ctx_tab.py
```diff
@@ -0,0 +1,286 @@
...
+            global ain,aout,dig
+            ain=ctx.getAnalogIn()
+            aout=ctx.getAnalogOut()
+            dig=ctx.getDigital()
+            #ps=ctx.getPowerSupply()
+            #trig=ain.getTrigger()
+            ctx_message.set("Context created.Connected to device with serial no.:")
+            ctx_serialno.set(str(ctx.getSerialNumber()))
+
+
```
commodo: nitpick: too many lines here

tools/python/m2k-osc/ain_tab.py
cristina-suteu: only tried python3 , so yes it works, but it probably works with python2 as well, haven't tried

tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,199 @@
...
+        self.linech0,=fig_subplot.plot(x,y_ch0)
+        self.linech1,=fig_subplot.plot(x,y_ch1)
+        self.canvas=FigureCanvasTkAgg(fig,master=self)
+        self.canvas.draw()
+        self.canvas.get_tk_widget().grid(row=11,column=0,columnspan=4)
+    def enable_ch0(self):
+        self.get_ain()
+        if ain_ch0_var.get() ==1:
+            ain.enableChannel(0,True)
+            print(ain.getRange(0))
```
commodo: this print looks like debug
intended or forgotten?

tools/python/m2k-osc/ain_tab.py
cristina-suteu: it's supposed to be stand-alone app to showcase functions of the libm2k library

tools/python/m2k-osc/aout_tab.py
```diff
@@ -0,0 +1,242 @@
+import matplotlib
+import numpy as np
+import tkinter as tk
+from tkinter import ttk
+import matplotlib.figure as  fg
+import libm2k
+from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg, NavigationToolbar2Tk
```
commodo: are all imports required?
or can some of them be reduced?
same question for globals

tools/python/m2k-osc/m2k-osc.py
```diff
@@ -0,0 +1,69 @@
...
+        tab_parent = ttk.Notebook()
+        self.tab_parent = tab_parent
+        tab_ctx = ctx_tab_frame(tab_parent, self)
+        tab_parent.add(tab_ctx,text='Context' )
+
+        tab_parent.grid(row=0,column=0)
+        tab_parent.grid_columnconfigure(0,weight=1)
+        tab_parent.grid_rowconfigure(0,weight=1)
+
+        """
```
commodo: for the final version, it's a good idea to remove commented code;
since this is a draft, it's not a problem;

tools/python/m2k-osc/aout_tab.py
cristina-suteu: imports and globals reduced

tools/python/m2k-osc/m2k-osc.py
cristina-suteu: commented sections removed

tools/python/m2k-osc/ain_tab.py
cristina-suteu: forgotten debug print. gone now

tools/python/m2k-osc/ain_tab.py
cristina-suteu: this is no longer an issue, nor does this code section exist in the current version

examples/analog/main.cpp
```diff
@@ -54,7 +54,6 @@ int main(int argc, char* argv[])
 	trig->setAnalogDelay(0);
```
commodo: this commit looks funky
it's only the `Signed-off-by` tag
maybe squash everything into the first one?

examples/analog/main.cpp
```diff
@@ -83,12 +82,12 @@ int main(int argc, char* argv[])
 	{
 		cout<<val<<endl;
 	}
-	cout << "CHANNEL2"<<endl;
+/*	cout << "CHANNEL2"<<endl;
```
commodo: this code looks commented out

commodo:
what is ` tools/python/libm2k ` ?

include/libm2k/config.hpp
```diff
@@ -0,0 +1,7 @@
+#ifndef CONFIG_HPP_CMAKEIN
+#define CONFIG_HPP_CMAKEIN
+
+#define PROJECT_VERSION "0.1"
+#define PROJECT_VERSION_GIT "1b12b79"
+
+#endif //CONFIG_HPP_CMAKEIN
```
commodo: is this file committed by accident?

tools/python/m2k-osc/ain_tab.py
```diff
@@ -0,0 +1,169 @@
...
+        self.calibscale_ch0.set(str(ain.getCalibscale(0)))
+        self.calibscale_ch1.set(str(ain.getCalibscale(1)))
+        self.vertoffset_ch0.set(str(ain.getVerticalOffset(0)))
+        self.vertoffset_ch1.set(str(ain.getVerticalOffset(1)))
+        self.rangelimits_ch0.set(str(ain.getRangeLimits(ain.getRange(0))))
+        self.rangelimits_ch1.set(str(ain.getRangeLimits(ain.getRange(1))))
+        self.hystrange_ch0.set(str(ain.getHysteresisRange(0)))
+        self.hystrange_ch1.set(str(ain.getHysteresisRange(1)))
+        self.sclfact_ch0.set(str(ain.getScalingFactor(0)))
+        self.sclfact_ch1.set(str(ain.getScalingFactor(1)))
```
commodo: i'm wondering if it would be a good idea to reverse the logic on this ch0 & ch1 attributes;
maybe group them somehow [a class or array, or hashtable]
then you could have some way like  `self.ch0.get_values()` & `self.ch1.get_values()`
you could also construct the ch0 & ch1 members with a single constructor;
`self.ch0 = self.init_ch(0)` or something like that;

i'm not saying this should be the exact code,but something along these lines would look more compact;

tools/python/m2k-osc/analog_custom_signal.csv
```diff
@@ -0,0 +1,37 @@
+0,
```
commodo: is this CSV file intended or committed by accident?

include/libm2k/config.hpp
cristina-suteu: definitely . apologies

examples/analog/main.cpp
cristina-suteu: this file is also committed by accident, sorry

tools/python/m2k-osc/analog_custom_signal.csv
cristina-suteu: this one is intended, you load the file in the tkinter app and push the values to an analog output channel. there's also a similar file to load and push data to the digital channels. they both serve as example files

PR #82, opened 2019-12-13
Documentation: Some minor changes to the Doxygen documentation.

AlexandraTrifan:
This makes the documentation for enums and structures more readable.

### DIFF
--- a/include/libm2k/enums.hpp
+++ b/include/libm2k/enums.hpp
@@ -63,8 +63,8 @@ namespace libm2k {
 
 
 	/**
-	* @enum M2K_TRIGGER_CONDITION_ANALOG for the analog side.
-	* @brief Condition of triggering.
+	* @enum M2K_TRIGGER_CONDITION_ANALOG for the analog side
+	* @brief Condition of triggering
 	*
 	*/
 	enum M2K_TRIGGER_CONDITION_ANALOG {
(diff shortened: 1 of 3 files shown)


include/libm2k/enums.hpp
```diff
@@ -111,50 +109,42 @@ namespace libm2k {
...
-		SRC_NONE = 2
+		SRC_TRIGGER_IN = 0, ///< SRC_TRIGGER_IN - trigger events on the TI(trigger in) pin trigger the DigitalIn interface
+		SRC_ANALOG_IN = 1, ///< SRC_ANALOG_IN - trigger events on the AnalogIn interface trigger the DigitalIn interface
+		SRC_NONE = 2 ///< SRC_NONE - trigger events on the DigitalIn are conditioned by the internal digital trigger structure
 	};
 
 
 	/**
 	 * @enum M2K_TRIGGER_OUT_SELECT
-	 *	SELECT_TRIGGER_IN - forwards trigger events from TI pin(trigger in);
```
commodo: curios: why not leave the docs for these values here?
would they be useful in the docs?

AlexandraTrifan: Doxygen generates a nicer format if we document this using "///< " instead of just writing them down in the "@brief" or "@enum" part.

commodo: ack
thanks for the info

PR #86, opened 2019-12-18
Communication(2): SPI

teoperisanu:
Add SPI tools: bitbanging and CLI.


commodo:
Would be nice to have some commit descriptions

teoperisanu:
> Would be nice to have some commit descriptions

Added commit descriptions.

PR #87, opened 2019-12-18
Communication(3): I2C

teoperisanu:
Add I2C tools: bitbanging and CLI.

### DIFF
--- a/tools/m2kcli/commands/communication/i2c.cpp
+++ b/tools/m2kcli/commands/communication/i2c.cpp
@@ -1,0 +1,169 @@
+/*
+ * Copyright 2019 Analog Devices, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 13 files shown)


tools/m2kcli/commands/communication/i2c.cpp
```diff
@@ -0,0 +1,189 @@
...
+				     "           [-h | --help]\n"
+				     "           [-i | --init frequency=<value> address=<value> scl=<index> sda=<index>]\n"
+				     "           [-w | --write data=<value>,... option=<value> [write_only=<value>]]\n"
+				     "           [-r | --read bytes_number=<value> option=<value>]\n"
+				     "\n"
+				     "Positional arguments:\n"
+				     "  uri                   describe the context location \n"
+				     "Optional arguments:\n"
+				     "  -h, --help            show this help message and exit\n"
+				     "  -i, --init [frequency=<value> address=<value> scl=<index> sda=<index> [write_only=<value>]]\n"
```
AlexandraTrifan: If we do something like: `m2kcli i2c ip:192.168.2.1 -i frequency=100000 address=0x48 scl=0 sda=1 write_only=1 -r bytes_number=1 option=1` will it crash to tell you something is wrong?

teoperisanu: Validation added in commit cb75049.

PR #91, opened 2019-12-18
Communication(5): UART

teoperisanu:
Add UART tools: bitbanging and CLI.

### DIFF
--- a/tools/m2kcli/main.cpp
+++ b/tools/m2kcli/main.cpp
@@ -27,6 +27,7 @@
 #include "commands/analog/power_supply.h"
 #include "commands/communication/spi.h"
 #include "commands/communication/i2c.h"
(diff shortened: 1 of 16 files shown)


tools/m2kcli/main.cpp
```diff
@@ -43,7 +44,7 @@ static const char *const helpMessage = "Usage:\n"
 				       "    analog-in           control the analogical input component\n"
 				       "    analog-out          control the analogical output component\n"
 				       "    digital             control the digital component\n"
-				       "    power-supply        control the power supply\n";
```
AlexandraTrifan: Is this a mistake? Power supply should be removed?

teoperisanu: Yes, there was a mistake. It was modified in commit 12f433b.

PR #94, opened 2019-12-19
Travis CI: Add build jobs for OSX builds

AlexandraTrifan:
Add build jobs for OSX builds. These create pkgs which can be published in a release.
Also, Python wheels are created in order to install Python bindings.


.travis.yml
```diff
@@ -19,6 +19,22 @@ matrix:
...
+      os: osx
+      osx_image: xcode10.1
+      env:
+        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
+    - compiler: "gcc"
+      os: osx
+      osx_image: xcode11
+      env:
+        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
+
```
commodo: nitpick: extra line

.travis.yml
```diff
@@ -19,6 +19,22 @@ matrix:
       env:
         - OS_TYPE=doxygen
       addons: { apt: { packages: ["cmake", "graphviz"] } }
+    - compiler: "gcc"
+      os: osx
+      osx_image: xcode9.2
+      env:
+        - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
```
commodo: curios: are we planning to store stuff on a server?
this env-var `EXTRA_SSH` is used in the upload scripts;
depending on how/if this is done, this is needed or not;

also, I would drop xcode9.2: it's really old;
xcode10 is old-ish;

at the rate Apple is updating Xcode, we might see an Xcode 12 pretty soon

CMakeLists.txt
```diff
@@ -19,11 +19,11 @@
 #
 
 cmake_minimum_required(VERSION 3.1.3)
-project(libm2k LANGUAGES CXX VERSION 0.1.0)
```
commodo: a commit desc would be nice

commodo:
rest looks good

.travis.yml
AlexandraTrifan: We'll have to store the artifacts somewhere. If it's not a problem, we can leave the "extra_ssh" there for now.

.travis.yml
AlexandraTrifan: removed

.travis.yml
commodo: Fine to leave it

PR #97, opened 2020-01-06
Travis CI:  Fix Python installation on OSX

AlexandraTrifan:
Drop the upgrade step for brew packages.
Upgrading to the latest version of some packages can result in undefined behaviour so it's better if we install the package if it does not exist.

### DIFF
--- a/CI/travis/before_install_darwin
+++ b/CI/travis/before_install_darwin
@@ -2,7 +2,7 @@
 
 . CI/travis/lib.sh
 
-brew_install_or_upgrade cmake doxygen libusb libxml2 swig curl python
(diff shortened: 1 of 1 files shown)


CI/travis/before_install_darwin
```diff
@@ -2,6 +2,7 @@
 
 . CI/travis/lib.sh
 
+brew unlink python@2
```
commodo: this looks a bit hacky;

curios: what is the error that this fixes?

AlexandraTrifan: In the latest build python upgrade from 3.5 to 3.6 fails because python2 is already linked.
Will modify this to install packages, not to install and upgrade them.

PR #101, opened 2020-01-10
Travis: Deploy packages to Github release page

AlexandraTrifan:
- This PR fixes some issues related to the OSX build (we don't want to build a framework by default). 
Some small changes were needed in order to create the .pkg and .tar.gz correctly.
- On Linux, we are building a .deb (for the base library) and a python .deb (only with python bindings) for every distribution on Travis-CI.
- Add code to deploy packages on the github release page for every newly created tag.
(2 .deb files and .tar.gz for Linux, and 2 .pkgs and 2 .tar.gz for OSX).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,11 @@ env:
 matrix:
   include:
     - os: linux
+      dist: bionic
+      python: 3.6
       env:
-        - OS_TYPE=ubuntu_docker
-        - OS_VERSION=bionic
+        - PYTHON=python PIP=pip
+        - OS_TYPE=ubuntu
     - os: linux
       dist: xenial
       env:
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -15,9 +15,11 @@ env:
 matrix:
   include:
     - os: linux
+      dist: bionic
+      python: 3.6
       env:
-        - OS_TYPE=ubuntu_docker
-        - OS_VERSION=bionic
+        - PYTHON=python PIP=pip
+        - OS_TYPE=ubuntu
```
commodo: this means no docker?
does Travis-CI have `dist: bionic` ?

commodo: seems so

commodo:
the changes look a bit messy;
but if they work, it's fine with me

.travis.yml
AlexandraTrifan: yes, they have `dist:bionic `

PR #103, opened 2020-01-16
CMakeLists.txt: Do not include rpath in the build tree.

AlexandraTrifan:
If the variable is set to False, the resulting .so/.dylib will have
the build folder added as rpath. This broke the Python bindings in
some scenarios.


dNechita:
Doesn't this change break stuff in case one wants to run things local?

PR #120, opened 2020-03-03
Calibration: Add handling for calibration data persistence.

teoperisanu:
Add the m2k-adc device as a member of M2kAnalogIn.
Save/load the calibration data from registers: offset, gain.
Save/load the vertical offset.

### DIFF
--- a/src/analog/m2kanalogin_impl.cpp
+++ b/src/analog/m2kanalogin_impl.cpp
@@ -50,6 +50,8 @@ M2kAnalogInImpl::M2kAnalogInImpl(iio_context * ctx, std::string adc_dev, bool sy
 	m_filter_compensation_table[1E4] = 1.20;
 	m_filter_compensation_table[1E3] = 1.26;
 
+	// calibbias attribute is only available in firmware versions newer than 0.26
+	m_calibbias_available = (Utils::compareVersions(Utils::getFirmwareVersion(ctx), "v0.26") > 0);
 	m_samplerate = 1E8;
 
 	for (unsigned int i = 0; i < getNbChannels(); i++) {
@@ -64,7 +66,6 @@ M2kAnalogInImpl::M2kAnalogInImpl(iio_context * ctx, std::string adc_dev, bool sy
 	if (sync) {
 		syncDevice();
 	}
-
 }
 
 M2kAnalogInImpl::~M2kAnalogInImpl() = default;
@@ -94,13 +95,10 @@ void M2kAnalogInImpl::init()
 		auto ch = static_cast<ANALOG_IN_CHANNEL>(i);
 
 		m_trigger->setAnalogMode(ch, ALWAYS);
-
 		setRange(ch, PLUS_MINUS_25V);
-		m_adc_calib_offset.at(i) = 2048;
-		m_adc_calib_gain.at(i) = 1;
-		m_adc_hw_vert_offset.at(i) = 0;
-                m_adc_hw_offset_raw.at(i) = 2048;
-		m_trigger->setCalibParameters(i, getScalingFactor(i), m_adc_hw_vert_offset.at(i));
+		setAdcCalibOffset(ch, 2048);
+		setAdcCalibGain(ch, 1);
+		setVerticalOffset(ch, 0);
 	}
 	setKernelBuffersCount(4);
 }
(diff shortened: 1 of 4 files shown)


src/analog/m2kanalogin_impl.cpp
```diff
@@ -105,55 +91,60 @@ void M2kAnalogInImpl::init()
...
-		m_adc_calib_gain.at(i) = 1;
-		m_adc_hw_vert_offset.at(i) = 0;
-		m_adc_hw_vert_offset_raw.at(i) = 2048;
+		setAdcCalibOffset(ch, 2048);
+		setAdcCalibGain(ch, 1);
+		setVerticalOffset(ch, 0);
 		m_trigger->setCalibParameters(i, getScalingFactor(i), m_adc_hw_vert_offset.at(i));
 	}
-	setKernelBuffersCount(1);
+	setKernelBuffersCount(4);
```
DanielGuramulta: I think this value `4` could be stored in a member variable, or static, maybe called `default_kernel_buffers` or w/e the naming convention is for the project.

src/analog/m2kanalogin_impl.cpp
```diff
@@ -105,55 +91,60 @@ void M2kAnalogInImpl::init()
...
-	m_adc_calib_gain[channel] = gain;
+
+	m_adc_calib_gain.at(channel) = m_m2k_adc->setDoubleValue(channel, gain, "calibscale");
 	m_trigger->setCalibParameters(channel, getScalingFactor(channel), m_adc_hw_vert_offset.at(channel));
 }
 
 void M2kAnalogInImpl::setAdcCalibOffset(ANALOG_IN_CHANNEL channel, int raw_offset)
 {
-	m_adc_calib_offset[channel] = raw_offset;
+	int rawVertOffset = m_adc_hw_offset_raw.at(channel) - m_adc_calib_offset.at(channel);
```
DanielGuramulta: This int can be made `const` as it is never modified.

src/analog/m2kanalogin_impl.cpp
```diff
@@ -105,55 +91,60 @@ void M2kAnalogInImpl::init()
...
+	m_adc_calib_offset.at(channel) = m_m2k_adc->setLongValue(channel, raw_offset, "calibbias", false);
+
+	int hw_offset_raw = rawVertOffset + m_adc_calib_offset.at(channel);
```
DanielGuramulta: Same as above.

src/analog/m2kanalogin_impl.cpp
```diff
@@ -510,82 +500,69 @@ std::vector<std::pair<std::string, std::pair<double, double>>> M2kAnalogInImpl::
...
-	double gain = 1.3;
-	double vref = 1.2;
-	int channel_number = 2 + channel;
-	double hw_range_gain = getValueForRange(m_input_range.at(channel));
-	int raw_vert_offset = (int) (vertOffset * (1 << 12) * hw_range_gain *
-				     gain / 2.693 / vref) + m_adc_calib_offset.at(channel);
 
-	m_ad5625_dev->setLongValue(channel_number, raw_vert_offset, "raw", true);
-	m_adc_hw_vert_offset_raw.at(channel) = raw_vert_offset;
+	int raw_vert_offset = convertVoltsToRawVerticalOffset(channel, vertOffset);
```
DanielGuramulta: This value can be `const`.

src/analog/m2kanalogin_impl.cpp
```diff
@@ -510,82 +500,69 @@ std::vector<std::pair<std::string, std::pair<double, double>>> M2kAnalogInImpl::
...
 	m_adc_hw_vert_offset.at(channel) = vertOffset;
-
+	int hw_offset_raw = raw_vert_offset + m_adc_calib_offset.at(channel);
```
DanielGuramulta: Same as above.

src/analog/m2kanalogin_impl.cpp
```diff
@@ -510,82 +500,69 @@ std::vector<std::pair<std::string, std::pair<double, double>>> M2kAnalogInImpl::
...
-	m_trigger->setCalibParameters(0, getScalingFactor(0), m_adc_hw_vert_offset.at(0));
-	m_trigger->setCalibParameters(1, getScalingFactor(1), m_adc_hw_vert_offset.at(1));
+	m_samplerate = m_m2k_adc->setLongValue((int)samplerate, "sampling_frequency");
+	m_trigger->setCalibParameters(chn_idx, getScalingFactor(chn_idx), m_adc_hw_vert_offset.at(chn_idx));return m_samplerate;
 }
 
 double M2kAnalogInImpl::getFilterCompensation(double samplerate)
 {
-	double compensation = 0.0;
+	double compensation;
```
DanielGuramulta: Is there any reason to not have this variable initialized here? http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-always

src/analog/m2kanalogin_impl.cpp
```diff
@@ -605,37 +582,82 @@ double M2kAnalogInImpl::getValueForRange(M2K_RANGE range)
...
+	m_m2k_adc->convertChannelHostFormat(chn_idx, avg, src, false);
+}
+
+int M2kAnalogInImpl::convertVoltsToRawVerticalOffset(ANALOG_IN_CHANNEL channel, double vertOffset)
+{
+	double gain = 1.3;
+	double vref = 1.2;
+	double hw_range_gain = getValueForRange(m_input_range.at(channel));
+	int raw_vert_offset = (int) (vertOffset * (1u << 12u) * hw_range_gain * gain / 2.693 / vref);
+	return raw_vert_offset;
```
DanielGuramulta: I think you can directly return the computation here without the need for a temporary variable `raw_vert_offset`.
Also c++ style casts should be preferred instead of the C ones.

src/analog/m2kanalogin_impl.cpp
```diff
@@ -605,37 +582,82 @@ double M2kAnalogInImpl::getValueForRange(M2K_RANGE range)
...
+}
+
+double M2kAnalogInImpl::convertRawToVoltsVerticalOffset(ANALOG_IN_CHANNEL channel, int rawVertOffset)
+{
+	double gain = 1.3;
+	double vref = 1.2;
+	double hw_range_gain = getValueForRange(m_input_range.at(channel));
+	double vert_offset = rawVertOffset * 2.693 * vref / ((1u << 12u) * hw_range_gain * gain);
```
DanielGuramulta: Same as above, w/o the cast comment.

src/analog/m2kanalogin_impl.hpp
```diff
@@ -147,6 +151,9 @@ class M2kAnalogInImpl : public M2kAnalogIn, public libm2k::utils::DeviceIn
 	const double *getSamplesInterleaved(unsigned int nb_samples, bool processed = false);
 
 	double processSample(int16_t sample, unsigned int channel);
+
+	int convertVoltsToRawVerticalOffset(ANALOG_IN_CHANNEL channel, double vertOffset);
```
DanielGuramulta: Shouldn't these methods be `const`? It seems to me that they won't affect the internal state of the `m2k_analogin_impl` object. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-fct

src/analog/m2kanalogin_impl.cpp
```diff
@@ -62,14 +49,15 @@ M2kAnalogInImpl::M2kAnalogInImpl(iio_context * ctx, std::string adc_dev, bool sy
 	m_filter_compensation_table[1E4] = 1.20;
 	m_filter_compensation_table[1E3] = 1.26;
 
-	m_samplerate = 1E8;
+	m_samplerate = m_m2k_adc->getDoubleValue("sampling_frequency");
```
AlexandraTrifan: Shouldn't we do this only in syncDevice() (so if we choose not to sync it, we just have the default values)?

src/analog/m2kanalogin_impl.cpp
```diff
@@ -510,82 +500,69 @@ std::vector<std::pair<std::string, std::pair<double, double>>> M2kAnalogInImpl::
...
+	return m_samplerate;
 }
 
 double M2kAnalogInImpl::setSampleRate(unsigned int chn_idx, double samplerate)
 {
-	return setDoubleValue(chn_idx, samplerate, "sampling_frequency");
-	m_trigger->setCalibParameters(0, getScalingFactor(0), m_adc_hw_vert_offset.at(0));
-	m_trigger->setCalibParameters(1, getScalingFactor(1), m_adc_hw_vert_offset.at(1));
+	m_samplerate = m_m2k_adc->setLongValue((int)samplerate, "sampling_frequency");
+	m_trigger->setCalibParameters(chn_idx, getScalingFactor(chn_idx), m_adc_hw_vert_offset.at(chn_idx));return m_samplerate;
```
AlexandraTrifan: Missing a "\n"?

src/analog/m2kanalogin_impl.cpp
teoperisanu: Done.

src/analog/m2kanalogin_impl.cpp
teoperisanu: Done.

src/analog/m2kanalogin_impl.cpp
teoperisanu: Done.

src/analog/m2kanalogin_impl.cpp
teoperisanu: Done.

src/analog/m2kanalogin_impl.cpp
teoperisanu: The modification was made.

src/analog/m2kanalogin_impl.cpp
teoperisanu: Done.

src/analog/m2kanalogin_impl.cpp
teoperisanu: Right, my mistake.

PR #133, opened 2020-04-23
M2k: Add a reset method to bring the ADALM2000 to a default state.

AlexandraTrifan:


teoperisanu:
Should resetting method destroy the buffers?

PR #136, opened 2020-04-29
M2K and Generic DAC fixes.

AlexandraTrifan:
- Fix the cyclic attribute for Generic Analog Out.
- Add a channel name getter for M2k Analog Out.

### DIFF
--- a/src/analog/m2kanalogout_impl.cpp
+++ b/src/analog/m2kanalogout_impl.cpp
@@ -23,6 +23,7 @@
 #include "utils/deviceout.hpp"
 #include <libm2k/m2kexceptions.hpp>
 #include <libm2k/utils/utils.hpp>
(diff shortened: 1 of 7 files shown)


src/analog/m2kanalogout_impl.cpp
```diff
@@ -598,3 +599,13 @@ unsigned int M2kAnalogOutImpl::getNbChannels()
 {
 	return m_dac_devices.size();
 }
+
+string M2kAnalogOutImpl::getChannelName(unsigned int channel)
+{
+	std::string name = "";
+	auto chn = getDacDevice(channel)->getChannel(0, true);
+	if (chn) {
+		name = chn->getName();
```
teoperisanu, line 608: The name of the iio_channel is empty. I think you want to retrieve its id (e.g. voltage0).

Issue #89, opened 2019-12-18
Save/Load calibration in Context


adisuciu:
Create Calibration class which contains all calibration values.These values can be loaded/saved at runtime by the user or by calling calibrateADC/calibrateDAC

PR #151, opened 2020-05-19
Calibrate using predefined values

teoperisanu:
An ini file with predefined calibration values can be generated or manually written. All its content will be written inside a context attribute at boot time if the file is added inside m2k driver. Then a fast calibration can be performed by using calibrateFromContext() method.
The purpose of this calibration type is to automate and to make the calibration
process more efficient.
This enhancement is described in #89.

### DIFF
--- a/bindings/python/calibration/README.md
+++ b/bindings/python/calibration/README.md
@@ -1,0 +1,47 @@
+# ADALM2000 calibrated using predefined values
+
+## Generating your own calibration file
+A calibration file is a file used at boot time to write inside the
+context some predefined values for the calibration process. The values
+will be stored until they will be overwritten.
+
+This file must be called 'm2k-calib-temp-lut.ini' and it should contain
+only a row having the following format:
+
+cal,temp_lut=[\<temperature>,<adc_offset1>,<adc_offset2>,<adc_gain1> <adc_gain2>,
+<dac_offset1>,<dac_offset2>,<dac_gain1>,<dac_gain2> [...]]
+
+### Command
+    generate_temperature_calib_lut <uri> [-h] [-t max_temp] [-T min] [-v nb_values] [-f path] [-a]
+
+### Examples
+Create the ini file 'm2k-calib-temp-lut.ini'. The process can be stopped by
+pressing 'CTRL + C', otherwise the process will stop when the temperature
+of the board rises up to 75 °C or after 30 minutes
+
+    python3 generate_temperature_calib_lut.py ip:192.168.2.1
+
+Create the ini file 'm2k-calib-temp-lut.ini'. The process can be stopped by
+pressing 'CTRL + C' or when the temperature of the board rises up to 54 °C
+
+    python3 generate_temperature_calib_lut.py auto -t 54
+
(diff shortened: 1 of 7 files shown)


adisuciu:
I think the .py file should be renamed to "generate_temperature_calib_lut.py" or something similar

src/m2k_impl.cpp
```diff
@@ -400,3 +402,71 @@ bool M2kImpl::getLed()
 	}
 	return on;
 }
+
+std::vector<double> M2kImpl::readCalibrationLine(double temperature)
```
adisuciu, line 406: This method should be refactored to show what it does  - something like:

getCalibrationValuesFromLUT()

adisuciu:
Check if the .py file can trim the number of decimals without losing precision in IIO so we can fit more data in the 4k string limit.

src/m2k_impl.cpp
teoperisanu: Modified.

src/m2k_impl.cpp
```diff
@@ -517,3 +520,101 @@ bool M2kImpl::getLed()
...
+	}
+
+	auto splitValues = Utils::split(values, ",");
+
+	auto it = splitValues.begin();
+	while (it != splitValues.end()) {
+		try {
+			temperature = std::stod(*it);
+			auto parameters = std::make_shared<struct CALIBRATION_PARAMETERS>();
+			parameters->adc_offset_ch_1 = std::stoi(*(++it));
```
DanielGuramulta, line 543: you can use `std::next(it)` instead of the ++ operator https://en.cppreference.com/w/cpp/iterator/next

bindings/python/calibration/README.md
```diff
@@ -0,0 +1,46 @@
...
+Create the ini file 'm2k-calib-temp-lut.ini'. Extract 5 values from all
+computed calibration parameters.
+
+    python3 generate_temperature_calib_lut.py auto -t nb_values=5 max_temp=54
+
+## Calibrating using ini file
+The purpose of this calibration type is to automate and to make the calibration
+process more efficient. Please make sure to have the following requirments:
+* A valid calibration file. The file can be generated using the script mentioned above.
+* Firmware v0.26 on your board.
```
DanielGuramulta, line 37: Maybe add a link to the firmware

bindings/python/calibration/README.md
```diff
@@ -0,0 +1,46 @@
...
+* Head of master - libm2k
```
DanielGuramulta, line 38: Maybe prefer a release instead of master branch: available with libm2k version > x.y.z

DanielGuramulta (commented):
Looks good from my point of view

bindings/python/calibration/README.md
teoperisanu: Modified.

bindings/python/calibration/README.md
teoperisanu: Added.

src/m2k_impl.cpp
teoperisanu: Modified.

DanielGuramulta (approved):

PR #155, opened 2020-05-27
Fix trigger level computation and add missing features.

AlexandraTrifan:
-These new features were necessary for integrating libm2k into Scopy.
-The trigger level fix refers to a computation error done when converting volts to raw samples.

### DIFF
--- a/src/analog/m2kanalogin_impl.cpp
+++ b/src/analog/m2kanalogin_impl.cpp
@@ -37,6 +37,7 @@ using namespace std::placeholders;
 M2kAnalogInImpl::M2kAnalogInImpl(iio_context * ctx, std::string adc_dev, bool sync, M2kHardwareTrigger *trigger) :
 	M2kAnalogIn(),
 	m_need_processing(false),
+	m_max_samplerate(-1),
 	m_trigger(trigger)
 {
 	m_m2k_adc = make_shared<DeviceIn>(ctx, adc_dev);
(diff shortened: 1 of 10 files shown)


src/analog/m2kanalogin_impl.cpp
```diff
@@ -37,7 +37,8 @@ using namespace std::placeholders;
 M2kAnalogInImpl::M2kAnalogInImpl(iio_context * ctx, std::string adc_dev, bool sync, M2kHardwareTrigger *trigger) :
 	M2kAnalogIn(),
 	m_need_processing(false),
-	m_trigger(trigger)
+	m_trigger(trigger),
+	m_max_samplerate(-1)
```
teoperisanu, line 41: These initializations should be switched one with the other, in order to suppress the order warning.

src/analog/m2kanalogin_impl.hpp
```diff
@@ -123,11 +123,13 @@ class M2kAnalogInImpl : public M2kAnalogIn
 	void getSamples(std::vector<std::vector<double> > &data, unsigned int nb_samples);
 
 	std::string getChannelName(unsigned int channel);
+	double getMaximumSamplerate();
```
teoperisanu, line 126: missing override keyword

src/analog/m2kanalogout_impl.hpp
```diff
@@ -86,8 +86,10 @@ class M2kAnalogOutImpl : public M2kAnalogOut
 
 	unsigned int getNbChannels();
 	std::string getChannelName(unsigned int channel);
+	double getMaximumSamplerate(unsigned int chn_idx);
```
teoperisanu, line 89: missing override keyword

src/analog/m2kanalogout_impl.cpp
```diff
@@ -622,3 +623,12 @@ string M2kAnalogOutImpl::getChannelName(unsigned int channel)
 	}
 	return name;
 }
+
+double M2kAnalogOutImpl::getMaximumSamplerate(unsigned int chn_idx)
+{
+	if (m_max_samplerate[chn_idx] < 0) {
```
teoperisanu, line 629: I would recommend to add an index validation

PR #177, opened 2020-08-26
M2kException: Fix segmantation fault.

teoperisanu:


AlexandraTrifan (approved):

src/m2kexception.cpp
```diff
@@ -115,7 +110,7 @@ m2k_exception_builder &m2k_exception_builder::line(int lineNumber)
 
 m2k_exception_builder &m2k_exception_builder::file(std::string fileName)
```
DanielGuramulta, line 111: `const std::string &fileName`

src/m2khardwaretrigger_v0.24_impl.cpp
```diff
@@ -187,9 +199,13 @@ M2K_TRIGGER_SOURCE_DIGITAL M2kHardwareTriggerV024Impl::getDigitalSource() const
 
 M2K_TRIGGER_SOURCE_DIGITAL M2kHardwareTrigger::getDigitalSource() const
 {
-	throw std::runtime_error("M2kHardwareTrigger: "
-				 "the digital external source is not available on "
-				 "the current board; Check the firmware version.");
+	throw m2k_exception::make("M2kHardwareTrigger: "
```
DanielGuramulta, line 202: Why not use the throw macro? Will it be removed?

teoperisanu: The macro uses throw_exception function, which will throw the exception only if exceptions are enabled. In these functions (from m2khardwaretrigger) the usage of the macro will cause a compile error, because when the exceptions are disabled the functions won't return anything.

src/m2kexception.cpp
teoperisanu: Modified.

src/m2khardwaretrigger_v0.24_impl.cpp
DanielGuramulta: So with exceptions disabled this hardware trigger "things" will still throw exceptions? Wouldn't it be more consistent to throw an exception when exceptions are enabled and return `nullptr` when exceptions are disabled?

teoperisanu: Modified.

DanielGuramulta (approved):

PR #182, opened 2020-09-02
Context: Add extra functionalities.

teoperisanu:
Functionalities:
- retrieve the iio context
- get iio backend version

### DIFF
--- a/src/context_impl.cpp
+++ b/src/context_impl.cpp
@@ -363,6 +363,11 @@ const struct libm2k::IIO_CONTEXT_VERSION ContextImpl::getIioContextVersion()
         return iioContextVersion;
 }
 
+struct iio_context *ContextImpl::getIioContext()
+{
+        return m_context;
+}
+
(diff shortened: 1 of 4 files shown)


src/context_impl.cpp
```diff
@@ -353,6 +353,16 @@ void ContextImpl::initializeContextAttributes()
 	}
 }
 
+struct libm2k::IIO_CONTEXT_VERSION ContextImpl::getIioContextVersion()
```
DanielGuramulta, line 356: This method could be `const`. If there are issues due to `iio_context_get_version` not being const, one way of implementing this would be to have the `iioContextVersion` as a member of the `ContextImpl` class that would be set in the constructor. Then this method could simply return it.

adisuciu (approved):

src/context_impl.cpp
teoperisanu: Modified.

PR #184, opened 2020-09-03
Calibration: Call stopAcquisition after every getSamplesRaw

adisuciu:
Previously stopAcquisition was only after calibrating ADC and DAC
alltogether. This causes the calibration to use continous buffer values
between calibration steps. ADC read does about 22 reads (one for offset
20 for fine tuning the offset and another one for gain), while DAC
calibration makes use of the ADC to calibrate offset and gain with one
read each. The buffers were continous to some measure due to the kernel
buffer configuration. By stopping the acquistion - we destroy the buffers
and force a reacquistion. This should give out better calibration values.

### DIFF
--- a/src/m2kcalibration_impl.cpp
+++ b/src/m2kcalibration_impl.cpp
@@ -95,8 +95,15 @@ void M2kCalibrationImpl::setAdcInCalibMode()
 	// Make sure hardware triggers are disabled before calibrating
 	m_trigger0_mode = m_m2k_trigger->getAnalogMode(ANALOG_IN_CHANNEL_1);
 	m_trigger1_mode = m_m2k_trigger->getAnalogMode(ANALOG_IN_CHANNEL_2);
+
 	m_m2k_trigger->setAnalogMode(ANALOG_IN_CHANNEL_1, ALWAYS);
 	m_m2k_trigger->setAnalogMode(ANALOG_IN_CHANNEL_2, ALWAYS);
+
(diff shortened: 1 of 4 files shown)


teoperisanu (approved):

src/m2kcalibration_impl.cpp
```diff
@@ -419,12 +422,14 @@ bool M2kCalibrationImpl::fine_tune(size_t span, int16_t centerVal0, int16_t cent
...
 
 		ch_data = m_m2k_adc->getSamplesRaw(num_samples);
+		m_m2k_adc->stopAcquisition();
 		if (ch_data.size() == 0) {
 			goto out_cleanup;
 		}
 
 		averagesCh0[i] = std::abs(Utils::average(ch_data.at(0).data(), num_samples));
 		averagesCh1[i] = std::abs(Utils::average(ch_data.at(1).data(), num_samples));
+
```
DanielGuramulta, line 432: nitpick: extra line

src/m2kcalibration_impl.cpp
```diff
@@ -419,12 +422,14 @@ bool M2kCalibrationImpl::fine_tune(size_t span, int16_t centerVal0, int16_t cent
 		std::this_thread::sleep_for(std::chrono::milliseconds(5));
 
 		ch_data = m_m2k_adc->getSamplesRaw(num_samples);
+		m_m2k_adc->stopAcquisition();
```
DanielGuramulta, line 425: No new line after `stopAcquisition()`, in some places a new line after `stopAcquisition()` was introduced. Should be consistent.

AlexandraTrifan (approved):

Issue #183, opened 2020-09-03
Create logging system for libm2k


adisuciu:
Should we use a library or go custom ?
glog is used by the ToF project
glog - https://github.com/google/glog

What should we log ? 

- library version
- firmware version
- libiio version
- libusb version 
- hardware access
- refills
- events (calibration/set trigger/ etc)
- exceptions 
- Developer errors (asserts)

Where should we log ?

- stderr
- stdout
- file - can you set this in the API ?

PR #185, opened 2020-09-08
Add logging system.

teoperisanu:
Create a logging mechanism using glog library.
Implements #183.

To do:
- [x] add warnings  (buffer size ...)
- [x] add more info logs (libs version, calibration, kernel buffers ...)
- [x] fix appveyor


src/CMakeLists.txt
```diff
@@ -48,10 +48,16 @@ endif()
 
 
 #Link the required dependencies
-target_link_libraries(${PROJECT_NAME}
-	${IIO_LIBRARIES}
-        ${LOG4CPP_LIBRARies}
-)
+if (ENABLE_LOG)
+	target_link_libraries(${PROJECT_NAME}
```
DanielGuramulta, line 52: I think you can do the usual target_link_libraries with IIO as that is required and if ENABLE_LOG just add glog
```cmake
target_link_libraries(${PROJECT_NAME} ${IIO_LIBRARIES})

if (ENABLE_LOG) 
       target_link_libraries(${PROJECT_NAME} glog::glog)
endif()
```

src/m2k_impl.cpp
```diff
@@ -198,12 +198,18 @@ bool M2kImpl::resetCalibration()
...
+	bool calibrationResult = m_calibration->calibrateADC();
+	LIBM2K_LOG(INFO, "[END] Calibrate ADC");
+	return calibrationResult;
 }
 
 bool M2kImpl::calibrateDAC()
 {
-	return m_calibration->calibrateDAC();
+	LIBM2K_LOG(INFO, "[BEGIN] Calibrate DAC");
+	bool calibrationResult = m_calibration->calibrateDAC();
```
DanielGuramulta, line 210: const

src/utils/buffer.cpp
```diff
@@ -38,10 +39,11 @@ Buffer::Buffer(struct iio_device *dev) {
 		m_dev = nullptr;
 		THROW_M2K_EXCEPTION("Buffer: Device not found, so no buffer can be created", libm2k::EXC_INVALID_PARAMETER);
 	}
+	m_dev_name = iio_device_get_name(m_dev);
```
DanielGuramulta, line 42: can this be placed in the initialiser list? I assume m_dev is know and it is the iio_device passed in Buffer(..)

src/utils/buffer.cpp
```diff
@@ -78,6 +80,17 @@ void Buffer::initializeBuffer(unsigned int size, bool cyclic, bool output)
 				THROW_M2K_EXCEPTION("Buffer: Cannot create the RX buffer", libm2k::EXC_RUNTIME_ERROR, errno);
 			}
 		}
+		if (output) {
+                        LIBM2K_LOG(INFO,
+                                   libm2k::buildLoggingMessage({m_dev_name}, ("TX buffer created (" + std::to_string(size) +
+                                                                      " samples)").c_str()));
+		} else {
```
DanielGuramulta, line 87: Instead of the if else would something like this work in the build logging message
```cpp
libm2k::buildLoggingMessage({m_dev_name}, ((output ? "TX" : "RX") + "buffer created ...
```

DanielGuramulta (approved):
Looks good!

DanielGuramulta:
One thing to keep in mind is that you are searching for glog version 0.3.5. This should be specified somewhere and build instructions for glog v0.3.5 should be provided. I don't think that, for example, `apt` will provide this version of glog on ubuntu 18.04 or 20.04. You could remove the version if you are sure that on every platform that libm2k is supported, the glog found there through package managers will have all the features you're using here.

src/CMakeLists.txt
teoperisanu: Modified.

src/m2k_impl.cpp
teoperisanu: Modified.

src/utils/buffer.cpp
teoperisanu: I thinks is a better approach to access the device structure after validating its integrity.

src/utils/buffer.cpp
teoperisanu: Modified.

PR #186, opened 2020-09-08
FakeM2k: first implementation of fakem2k

adisuciu:
This commit implements a fake m2k device. It is based on the same m2k API.
The code is copied from the m2k implementation, but the hardware access
has  been replaced with some defaults. One needs to open an xml context
(also provided in the commit) using the uri "xml:<path_to_xml_file>" to use
the fake m2k device.

Future implementation:
1. implement behavior in the fake m2k device (data/loopbacks/samplerates
/gain modes/ triggers/ etc ... )
2. reuse code by using the same m2kimplementation files but use hardware
generics (Buffer/Device/etc) or FakeGenerics (their behavior could be
modeled in a separate behavior class)


teoperisanu:
A better approach is taken into consideration. This approach does not imply any changes neither in libm2k nor in libiio. The solution is an implementation for the ADI tinyiiod (https://github.com/analogdevicesinc/libtinyiiod) that simulates the behavior of m2k.

tfcollins:
> A better approach is taken into consideration. This approach does not imply any changes neither in libm2k nor in libiio. The solution is an implementation for the ADI tinyiiod (https://github.com/analogdevicesinc/libtinyiiod) that simulates the behavior of m2k.
@teoperisanu do we have any working (or somewhat working) implementation of this with libtinyiiod? I've been exploring different ways of mocking backends to test remote contexts.

teoperisanu:
> > A better approach is taken into consideration. This approach does not imply any changes neither in libm2k nor in libiio. The solution is an implementation for the ADI tinyiiod (https://github.com/analogdevicesinc/libtinyiiod) that simulates the behavior of m2k.
> 
> @teoperisanu do we have any working (or somewhat working) implementation of this with libtinyiiod? I've been exploring different ways of mocking backends to test remote contexts.

Yes, we have. Our first approach was to override the methods inside the iio_context , but this came with an extra step of assigning the ops structure to the xml context.
Currently we are implementing the logic using libtinyiiod and running the server locally. There are still some possibilities to be explored, but so far it seems to be the right way.

PR #188, opened 2020-09-15
Fixes

### LINKED ISSUE #196: Fix logger warnings
libm2k currently throws a lot of warnings from the logging piece. Please fix these.

DanielGuramulta:
Should clean a bunch of warnings. Some appear due to clang being more aggressive on what's enabled by default.
Also fixes #196.

https://github.com/analogdevicesinc/libm2k/blob/master/src/analog/m2kanalogin_impl.cpp#L474

```cpp
const short *M2kAnalogInImpl::getVoltageRawP()
{
	std::vector<short> avgs = getVoltageRaw();
	return avgs.data();
}
```

- [x] `warning: address of stack memory associated with local variable 'avgs' returned [-Wreturn-stack-address]` . This is undefined behaviour. Is the std::vector<short> a 1 element vector containing an average value?

Can this return a `short` value representing the average or a `short *` containing an array of averages maybe through a smart ptr?
```cpp
memcpy(return value, avgs.data(), ...)
```
The problem also appears in the `double` version of this method.

- [x] can I add -Wreturn-stack-address to be threated as an error?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -40,6 +40,22 @@ if (MSVC)
 	add_definitions("/wd4250 /wd4251 /wd4275")
 endif()
 
+if(NOT MSVC)
+	target_compile_options(${PROJECT_NAME}
+			PRIVATE
+			-Wall
+			-Wextra)
+	if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+		target_compile_options(${PROJECT_NAME}
+				PRIVATE
+				-Wreturn-stack-address)
+	elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+		target_compile_options(${PROJECT_NAME}
+				PRIVATE
+				-Wreturn-local-addr)
+	endif()
+endif()
+
 target_include_directories(${PROJECT_NAME}
     PUBLIC
         $<INSTALL_INTERFACE:include>
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -36,6 +36,9 @@ target_compile_definitions(${PROJECT_NAME} PRIVATE LIBM2K_EXPORTS)
 if (ENABLE_LOG)
 	target_compile_definitions(${PROJECT_NAME} PUBLIC LIBM2K_ENABLE_LOG)
 endif()
+if (MSVC)
+	add_definitions("/wd4250 /wd4251 /wd4275")
+endif()
 
 target_include_directories(${PROJECT_NAME}
     PUBLIC
--- a/src/analog/m2kanalogin_impl.hpp
+++ b/src/analog/m2kanalogin_impl.hpp
@@ -81,8 +81,18 @@ public:
 
 	std::pair<double, double> getHysteresisRange(ANALOG_IN_CHANNEL chn) override;
 
+	/**
+	 * The calibration offset is store in the same register with the vertical offset as a sum (CalibOff + VertOff)
+	 * The calibration offset is always represented as an integer raw value
+	 * The vertical offset is store as an integer raw value, but it can be converted to a voltage representation
+	 * There are several ways to set these coefficients
+	 * */
+	 // set the calibration offset - use the current vertical offset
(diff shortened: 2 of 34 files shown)


src/CMakeLists.txt
```diff
@@ -36,6 +36,16 @@ target_compile_definitions(${PROJECT_NAME} PRIVATE LIBM2K_EXPORTS)
 if (ENABLE_LOG)
 	target_compile_definitions(${PROJECT_NAME} PUBLIC LIBM2K_ENABLE_LOG)
 endif()
+if (MSVC)
+	add_definitions(
+			/wd4250
```
DanielGuramulta, line 41: maybe add a comment for each of this, on what they do: `disable warning "abc"`

DanielGuramulta:
If there are no more warnings that show up on the CI, -WError could be enforced to promote cleaner code and 0 warnings in future code changes. @AlexandraTrifan @teoperisanu  what do you think of this?

teoperisanu:
I think this is a good idea, but there are still some warnings that prevent this modification.

src/CMakeLists.txt
```diff
@@ -36,6 +36,16 @@ target_compile_definitions(${PROJECT_NAME} PRIVATE LIBM2K_EXPORTS)
...
+			/wd4251
+			/wd4275)
+endif()
+
+if(NOT MSVC)
+	target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wextra -Wreturn-local-addr)
```
DanielGuramulta, line 47: I think this should be `-Wreturn-stack-address` not `-Wreturn-local-addr`

teoperisanu: -Wreturn-local-addr is a gcc flag, while -Wreturn-stack-address is a clang one, both having the same meaning.

DanielGuramulta: Well then for gcc/clang the corresponding flag should be used. One way of doing this is multiple if branches where we add specific flags for clang and for gcc having the common ones in one place (`-Wall, -Wextra`) or using generator expressions
```cmake
target_compile_options(${PROJECT_NAME}
    PRIVATE
        -Wall
        -Wextra
        $<$<LANG_COMPILER_ID:GNU>: -Wreturn-local-addr>
        $<$<LANG_COMPILER_ID:Clang, AppleClang>: -Wreturn-stack-address>
)
```
Hope i got the syntax right.

DanielGuramulta: If we are compiling with clang -Wreturn-local-addr will have no effects, and viceversa gcc with -Wreturn-stack-address

teoperisanu: Modified.

src/CMakeLists.txt
teoperisanu: Added.

DanielGuramulta:
There are also some warnings about uninitialised values coming from the csharp bindings. Those should be disabled also.
I think Wno-maybe-uninitialised does the job.

teoperisanu:
> There are also some warnings about uninitialised values coming from the csharp bindings. Those should be disabled also.
> I think Wno-maybe-uninitialised does the job.

Now the warning should be suppressed.

AlexandraTrifan (approved):

teoperisanu (approved):

src/analog/m2kanalogin_impl.hpp
```diff
@@ -83,6 +83,7 @@ class M2kAnalogInImpl : public M2kAnalogIn
 
 	void setAdcCalibOffset(ANALOG_IN_CHANNEL channel, int calib_offset);
 	void setAdcCalibOffset(ANALOG_IN_CHANNEL channel, int calib_offset, int vert_offset);
+	void setAdcCalibOffset(ANALOG_IN_CHANNEL channel, int calib_offset, double vert_offset);
```
AlexandraTrifan, line 86: should we add a comment to specify that one function is using the offset in Volts and the other one in Raw values? 
maybe it will help for further development

teoperisanu: I added some comments to clarify the usage of each function.

PR #191, opened 2020-09-25
AnalogIn & AnalogOut: Handle the exception raised when converting a string to double.

AlexandraTrifan:


DanielGuramulta (approved):
Looks good!

PR #194, opened 2020-10-09
Log all attributes

teoperisanu:
This functionality is useful to define the initial state of the board for eventual debugging processes.


src/context_impl.hpp
```diff
@@ -70,12 +70,12 @@ class ContextImpl : public virtual Context {
 	libm2k::analog::DMM* getDMM(std::string);
 	std::vector<libm2k::analog::DMM*> getAllDmm();
 
-
 	std::vector<std::string> getAvailableContextAttributes();
 	std::string getContextAttributeValue(std::string attr);
 	std::string getContextDescription();
 	std::string getSerialNumber();
 	std::unordered_set<std::string> getAllDevices();
+	void logAllAttributes() override;
```
DanielGuramulta, line 78: can this method be tagged `const`?

DanielGuramulta (approved):
Looks good!

src/context_impl.hpp
teoperisanu: I think that const for void function has no utility.

DanielGuramulta: it's not about the return type, it's about the const correctness of the api. Even if a method is void it can still alter the internal state of the object/ or not. Example `void clear()` of a `std::vector`. In our case here: does the logging of all the attributes of the context, alter its internal state in any way?

DanielGuramulta: Another example:
```cpp
class Whatever {
public:
  void inspect() const;   // This member promises NOT to change *this
  void mutate();          // This member function might change *this
};
```

AlexandraTrifan (approved):

src/context_impl.hpp
teoperisanu: Modified.

DanielGuramulta (approved):

PR #195, opened 2020-10-19
M2KCalibration: use calibration done flags for firmware 0.25 and earlier

adisuciu:
This is because for older firmware you cannot tell whether the voffset
has been calibrated or not. This causes an offset when using default values
even though isCalibrated returns true

### DIFF
--- a/src/m2k_impl.cpp
+++ b/src/m2k_impl.cpp
@@ -539,6 +539,12 @@ bool M2kImpl::getLed()
 
 bool M2kImpl::isCalibrated()
 {
+	int diff = Utils::compareVersions(m_firmware_version, "v0.25");
+	if (diff <= 0) {
+		// for FW 0.25 and earlier we cannot conclude if the adc/dac was calibrated due to missing calibbias attribute
+		return m_calibration->getAdcCalibrated() && m_calibration->getDacCalibrated();
+	}
+
 	for (unsigned int i = 0; i < 2; i++) {
 		if (m_calibration->getAdcOffset(i) != defaultOffset || m_calibration->getDacOffset(i) != defaultOffset) {
 			return true;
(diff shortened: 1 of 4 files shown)


src/m2kcalibration_impl.hpp
```diff
@@ -74,6 +74,10 @@ class M2kCalibrationImpl : public M2kCalibration
 	void setDacOffset(unsigned int chn, int offset);
 	void setAdcOffset(unsigned int chn, int offset);
 	void setAdcGain(unsigned int chn, double gain);
+
+	bool getAdcCalibrated() const;
```
DanielGuramulta, line 78: this methods can be declared as `override`

src/m2k_impl.cpp
```diff
@@ -539,6 +539,12 @@ bool M2kImpl::getLed()
 
 bool M2kImpl::isCalibrated()
 {
+	int diff = Utils::compareVersions(m_firmware_version, "v0.25");
+	if (diff <= 0) {
+		// for FW 0.25 and earlier we cannot conclude if the adc/dac was calibrated due to missing calibbias attribute
+		return m_calibration->getAdcCalibrated() && m_calibration->getAdcCalibrated();
```
DanielGuramulta, line 545: shouldn't this be 
```cpp
return m_calibration->getAdcCalibrated() && m_calibration->getDacCalibrated();
```

DanielGuramulta (changes requested):

src/m2k_impl.cpp
adisuciu: You are correct .. it is a typo

src/m2kcalibration_impl.hpp
adisuciu: fixed

DanielGuramulta (approved):
Looks good!

AlexandraTrifan (approved):

PR #197, opened 2020-10-23
CMakeLists: bindings no longer use -Wall and standard is set with CMa…

adisuciu:
…ke command


teoperisanu:
This issue should be fixed in #188.

teoperisanu:
This issue should be fixed in #188.

PR #202, opened 2020-11-03
M2kCalibration: Reduce the time spent on the calibration process.

AlexandraTrifan:
Fine tuning runs about 20 times. This creates the buffer, fills it and
destroys it right after. We reduce the amount of time spent here by
using the kernel buffers and the fact that the buffer is refilled instantly
after the buffer is created.

This reduces the calibration time with almost 1s.
Ran some tests to check calibration accuracy, results are the following:

> Test 1: Reset and get default values
2048 2048 1.0 1.0 2048 2048 0.718079 0.718079
CH0  0.028281165357529826
CH1  0.01082365587757312

>Test 2: Reading values after Scopy v1.2.0 did the calibration.
2011 2021 1.066284 1.075745 2072 2075 0.646606 0.639526
CH0  -0.005844470565028971
CH1  -0.0020493598085166514

>Test 1: Reset and get default values
2048 2048 1.0 1.0 2048 2048 0.718079 0.718079
CH0  0.02727925611781052
CH1  0.010535227460078176

>Test 3: Calibrate using the modified libm2k calibration and compare them with the previous results.
2010 2021 1.067749 1.076477 2074 2077 0.64624 0.638184
CH0  -0.0045085915787366265
CH1  0.002793201516793067

The difference between raw offsets of this version and the current master version are max 2. The readings for 0V output look good.
The test values are listed as follows:
ADC_CH0_Offset,  ADC_CH1_Offset,   ADC_CH0_Gain,     ADC_CH1_Gain,    DAC_CH0_Offset,  DAC_CH1_Offset,   DAC_CH0_Gain,     DAC_CH1_Gain

Signed-off-by: AlexandraTrifan <Alexandra.Trifan@analog.com>

### DIFF
--- a/include/libm2k/analog/m2kanalogin.hpp
+++ b/include/libm2k/analog/m2kanalogin.hpp
@@ -418,6 +418,12 @@ public:
 	 */
 	virtual void setKernelBuffersCount(unsigned int count) = 0;
 
+	/**
+	 * @brief Get the number of kernel buffers
+	 * @return the number of previously set kernel buffers (saved in this session)
+	 */
+	virtual unsigned int getKernelBuffersCount() const = 0;
+
 
 	/**
 	* @brief Get the hardware trigger handler
--- a/include/libm2k/analog/m2kanalogout.hpp
+++ b/include/libm2k/analog/m2kanalogout.hpp
@@ -392,6 +392,13 @@ public:
 	 */
 	virtual void setKernelBuffersCount(unsigned int chnIdx, unsigned int count) = 0;
 
+	/**
+	 * @brief Get the number of kernel buffers
+	 * @param chnIdx The index corresponding to the channel
+	 * @return the number of previously set kernel buffers (saved in this session)
+	 */
+	virtual unsigned int getKernelBuffersCount(unsigned int chnIdx) const = 0;
+
 
 	/**
 	 * @brief Convert the volts value of a sample into raw
--- a/src/analog/m2kanalogout_impl.cpp
+++ b/src/analog/m2kanalogout_impl.cpp
@@ -112,6 +112,27 @@ void M2kAnalogOutImpl::syncDevice()
 	LIBM2K_LOG(INFO, "[END] M2kAnalogOut sync");
 }
 
+void M2kAnalogOutImpl::loadNbKernelBuffers()
+{
+	if (m_dma_data_available) {
+		const unsigned int buffersize = 16;
+		for (unsigned int  chn = 0; chn < m_dac_devices.size(); chn++) {
+			auto ch_en = isChannelEnabled(chn);
+			getDacDevice(chn)->enableChannel(0, true, true);
+			getDacDevice(chn)->initializeBuffer(buffersize, false);
+			auto sampleSize = getDacDevice(chn)->getSampleSize();
+			auto unusedBufferSpace = m_dac_devices[chn]->getBufferLongValue("data_available");
+			m_nb_kernel_buffers[chn] = (unusedBufferSpace / buffersize) / sampleSize + 1;
+			getDacDevice(chn)->stop();
+			getDacDevice(chn)->enableChannel(0, ch_en, true);
+		}
+	} else {
+		for (unsigned int  chn = 0; chn < m_dac_devices.size(); chn++) {
+			m_nb_kernel_buffers[chn] = 4;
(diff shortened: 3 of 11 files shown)


src/m2kcalibration_impl.cpp
```diff
@@ -475,6 +504,7 @@ bool M2kCalibrationImpl::fine_tune(size_t span, int16_t centerVal0, int16_t cent
 	return ret;
 }
 
+
```
DanielGuramulta, line 507: nitpick: is this line required?

DanielGuramulta (approved):
Looks good!

include/libm2k/analog/m2kanalogin.hpp
```diff
@@ -418,6 +418,12 @@ class LIBM2K_API M2kAnalogIn
 	 */
 	virtual void setKernelBuffersCount(unsigned int count) = 0;
 
+	/**
+	 * @brief Get the number of kernel buffers
+	 * @return the number of kernel buffers
+	 */
+	virtual unsigned int getKernelBuffersCount() = 0;
```
teoperisanu, line 425: This method can be const.

include/libm2k/analog/m2kanalogout.hpp
```diff
@@ -392,6 +392,13 @@ class LIBM2K_API M2kAnalogOut
 	 */
 	virtual void setKernelBuffersCount(unsigned int chnIdx, unsigned int count) = 0;
 
+	/**
+	 * @brief Get the number of kernel buffers
+	 * @param chnIdx The index corresponding to the channel
+	 * @return the number of kernel buffers
+	 */
+	virtual unsigned int getKernelBuffersCount(unsigned int chnIdx) = 0;
```
teoperisanu, line 400: This method can be const.

teoperisanu (changes requested):

src/analog/m2kanalogout_impl.cpp
```diff
@@ -112,6 +113,27 @@ void M2kAnalogOutImpl::syncDevice()
 	LIBM2K_LOG(INFO, "[END] M2kAnalogOut sync");
 }
 
+void M2kAnalogOutImpl::loadNbKernelBuffers()
+{
+	if (m_dma_data_available) {
+		unsigned int buffersize = 16;
```
teoperisanu, line 119: const

src/m2kcalibration_impl.cpp
```diff
@@ -120,6 +120,11 @@ void M2kCalibrationImpl::setAdcInCalibMode()
 
 	m_m2k_adc->enableChannel(0, true);
 	m_m2k_adc->enableChannel(1, true);
+
+	m_m2k_adc->loadNbKernelBuffers();
+	m_adc_kernel_buffers = m_m2k_adc->getKernelBuffersCount();
+	std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
teoperisanu, line 126: This sleep is somehow useful because otherwise the application will crash. This is caused because of the device that did not have enough time to perform all previous operations, so iio_device_set_kernel_buffers_count will return the code -EBUSY.
Another possible workaround can be something like this:
`try_again:
	try {
		m_m2k_adc->setKernelBuffersCount(1);
	} catch (m2k_exception &e) {
		if (e.iioCode() == -EBUSY) {
			goto try_again;
		} else {
			THROW_M2K_EXCEPTION(e.what(), e.type(), e.iioCode());
		}
	}`

DanielGuramulta: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-goto
Better ditch the `goto`  statement and use a `for loop` or something else. I think this try while EBUSY approach is better than random sleeping `x`ms. One thing to be careful is that if you trow an exception in the else branch some methods in the calibration have a 'clean_up' part deleting some allocated arrays. One way of fixing that is using smart pointers.

adisuciu: I think sleeping is fine.
If calibration is done in another thread, other threads can catch-up. The setting of a kernel buffer with try/catch is convoluted and not obvious. I'd rather just put in a comment, sleep because setting this value is not "instant", or similar.

teoperisanu (approved):

PR #205, opened 2020-11-27
General: Replace "std:stod" with string streams

AlexandraTrifan:
Use imbue to associate a locale to the stream and stream buffer.

std::stod behaviour changes based on the system locale, we made these
changes to avoid wrong parameter conversion.

- [x] Verify: https://github.com/analogdevicesinc/libm2k/blob/master/bindings/python/calibration/generate_temperature_calib_lut.py
Check if the generated calibration parameters file depends on the system locale (42.5 degrees or 42,5 degrees affect the way values are loaded into the library)

Signed-off-by: AlexandraTrifan <Alexandra.Trifan@analog.com>

### DIFF
--- a/include/libm2k/utils/utils.hpp
+++ b/include/libm2k/utils/utils.hpp
@@ -71,6 +71,7 @@ namespace utils {
 		static std::vector<std::string> split(std::string, std::string);
 		static int compareVersions(std::string v1, std::string v2);
 		static bool compareNatural(const std::string &a, const std::string &b);
+		static double safeStod(const std::string& to_convert);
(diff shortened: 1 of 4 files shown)


teoperisanu:
There are still some stod calls left unchanged.(https://github.com/analogdevicesinc/libm2k/blob/8c71f01e2cbfd7a1958ec8852eaa6c1493df3129/src/m2k_impl.cpp#L574-L584).

AlexandraTrifan:
Added the required changes.

teoperisanu (approved):

PR #206, opened 2020-12-02
M2kDigital: Add missing m2k_exception.

### LINKED ISSUE #204: Replace 'exception_type' with 'm2k_exception'
In some places (such as in m2kdigital_impl.cpp) the old code - before m2k_exception was introduced - was left where the exception_type is catch. This leads to the loss of certain information (such as iio_code) from m2k_exception.

teoperisanu:
First try to catch m2k_exception, otherwise the iio code will be lost.

Fixes #204.


DanielGuramulta (approved):
Looks good!

PR #212, opened 2021-01-11
CMakeLists.txt: Add support for versions less than 3.8.2.

### LINKED ISSUE #9: CMake requirements incorrect
To use the cmake command SWIG_ADD_LIBRARY requires CMake 3.8, need to bump the required version or tweak the python bindings cmake file. Might want to check how easy 3.8 is to get on Ubuntu.

teoperisanu:
Fixes #9.


AlexandraTrifan (approved):
Looks good!

PR #241, opened 2021-06-17
Logic Analyser External Clocksource

cristina-suteu:
added setter and getter for clock source attribute
added Python external clock source example

### DIFF
--- a/include/libm2k/digital/m2kdigital.hpp
+++ b/include/libm2k/digital/m2kdigital.hpp
@@ -423,6 +423,22 @@ public:
 	 */
 	virtual void resetRateMux() = 0;
 
+	/**
+	 * @brief Set the clocksource
+	 * @param external - True to set source to external
+	 *		   - False to set source to internal
+	 *
+	 * @note Only available from firmware v0.26.
+	 */
+	virtual void setExternalClocksource(bool external) = 0;
+
+	/**
+	 * @brief Get the current clocksource
+	 * @return a string of the clocksource
+	 *
+	 * @note Only available from firmware v0.26.
+	 */
+	virtual std::string getClocksource() = 0;
 };
 }
 }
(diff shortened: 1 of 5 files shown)


src/digital/m2kdigital_impl.cpp
```diff
@@ -493,3 +493,23 @@ void M2kDigitalImpl::resetRateMux()
...
+			m_dev_generic->setStringValue("clocksource", "external");
+		else
+			m_dev_generic->setStringValue("clocksource", "internal");
+	}
+
+}
+
+std::string M2kDigitalImpl::getClocksource()
+{
+	if(m_dev_generic->hasGlobalAttribute("clocksource")){
```
teoperisanu, line 510: This function should return something in case clocksource is not defined or throw an exception (EXC_INVALID_FIRMWARE_VERSION).

include/libm2k/digital/m2kdigital.hpp
```diff
@@ -423,6 +423,22 @@ class LIBM2K_API M2kDigital
...
+	 */
+	virtual void setExternalClocksource(bool external) = 0;
+
+	/**
+	 * @brief Get the current clocksource
+	 * @return a string of the clocksource
+	 *
+	 * @note Only available from firmware v0.26.
+	 */
+	virtual std::string getClocksource() = 0;
```
teoperisanu, line 441: Because the setter uses boolean value, I suppose that the getter should work in the same way. Something like bool isExternClockSet().

src/digital/m2kdigital_impl.cpp
```diff
@@ -493,3 +493,38 @@ void M2kDigitalImpl::resetRateMux()
...
+			return true;
+		} else {
+			return false;
+		}
+	} else {
+		THROW_M2K_EXCEPTION("M2kDigital: "
+				    "There is no clocksource attribute on "
+				    "the current board; Check the firmware version.",
+				    libm2k::EXC_INVALID_FIRMWARE_VERSION);
+	}
```
teoperisanu, line 527: Add return false after throw, because in some cases THROW_M2K_EXCEPTION can have exceptions disabled.

teoperisanu (approved):

PR #244, opened 2021-06-19
udev add uaccess tag

maitbot:
The modern way to allow console users access to USB gadgets is the uaccess tag.
This patch adds the udev uaccess tag.
See https://wiki.debian.org/USB/GadgetSetup
and the lintian check: https://lintian.debian.org/tags/udev-rule-missing-uaccess


AlexandraTrifan:
Could you change the target branch for this PR to master?
We will soon create a new minor release, which contains some other fixes and will include this as well.

adisuciu:
Can you also please --signoff your commits ? 
Thanks,
Adrian

PR #245, opened 2021-06-19
cleaner release version

maitbot:
Improve version reporting for the case without a git hash,
such as a release version, or tarball package build.


AlexandraTrifan:
Could you change the target branch for this PR to master?
We will soon create a new minor release, which contains some other fixes and will include this as well.

PR #247, opened 2021-06-23
udev add uaccess tag

maitbot:
The modern way to allow console users access to USB gadgets is the uaccess tag.

Signed-off-by: A. Maitland Bottoms <bottoms@debian.org>


AlexandraTrifan (approved):
Looks good, thank you!

PR #280, opened 2021-09-28
DMM: Modify struct to include unit name and symbol as defined by the International System of Units.

AlexandraTrifan:
- [x] run tests on Windows for MSVC and C# bindings.

Signed-off-by: Alexandra Trifan <Alexandra.Trifan@analog.com>


adisuciu (approved):

adisuciu:
key_symbol = "kPa";
key_symbol = "mA";
I don't think these units should have prefixes

PR #319, opened 2022-08-28
Output values of channel 1 and 2 to file

mzaffalon:


mzaffalon:
I find it more useful to output to file instead of stdout. In this way I could see that write and read operations in the M2K are not synchronous.

AlexandraTrifan:
The examples are provided for guidance on how to use the libm2k API. Writing the acquired data to file is beyond the purpose of this example. 
The user can implement any preferred way to manipulate the data (whether it's written directly to a file, redirected from the console or any other way).

mzaffalon:
Currently, the data is output to console, which is in my opinion not more useful then writing it to file.

rgetz:
Data bring output to console- is sort of the "Unix way". https://en.m.wikipedia.org/wiki/Unix_philosophy

That way if can be piped into any tool you want. 

- Robin

PR #324, opened 2022-09-16
tests: Added tests that use the iio-emu emulator for ADALM2000

bindea-cristian:
Added a new section of tests that use the M2K emulator.

Signed-off-by: Cristian Bindea <cristian.bindea@analog.com>


AlexandraTrifan:
Can you add some details about the emulator tests and a simple example to the README page from the tests folder?
Shouldn't the tests also run the iio-emu device instead of assuming that the user started it before running the tests?

AlexandraTrifan (approved):

PR #345, opened 2023-09-27
M2kAnalogIn: sync calibration state with device attributes

Adrian-Stanea:
- Use device attributes in class instance for calibration at the SW level, samples should be sent uncalibrated. 
- Current firmware uses device attr to send calibrated samples by the HW. To simulate future firmware behaviour where samples are not calibrated we store the calibration result in the class instance, but don't write them to device attributes.


Adrian-Stanea:
Tested potential fix in old firmware. The solution is ambiguous since it leads to calibrate call that wont be reflected in the iio attributes, only in class instance. 
The change also requires a new calibration for each new instance since the values are not cached in iio.

PR #351, opened 2023-11-29
fw0.32/adc_calib: sync calibration state and use new attribute to decouple HW calibration

Adrian-Stanea:
- Sync class parameters: m_adc_calib_offset, m_adc_calib_gain with device attributes. These are used to calibrate the raw data to volts.
- In firmware version 0.32, the device attribute won't be used by the hardware, samples will be recieved uncalibrated. The calibration will be done in software.

- In new libm2k, use the calibration attr is used to disable setting the computed calibration values in HW.
- Note: - set calibration = 'true' (string vale): stops the HW calibration - set calibration = 'false' (string value): enables the HW calibration

### DIFF
--- a/src/analog/m2kanalogin_impl.cpp
+++ b/src/analog/m2kanalogin_impl.cpp
@@ -43,9 +43,15 @@ M2kAnalogInImpl::M2kAnalogInImpl(iio_context * ctx, std::string adc_dev, bool sy
 	m_trigger(trigger)
 {
 	LIBM2K_LOG(INFO, "[BEGIN] Initialize M2kAnalogIn");
+	firmware_version = Utils::getFirmwareVersion(ctx);
+
 	m_m2k_adc = make_shared<DeviceIn>(ctx, adc_dev);
 	m_m2k_fabric = make_shared<DeviceGeneric>(ctx, "m2k-fabric");
 	m_ad5625_dev = make_shared<DeviceGeneric>(ctx, "ad5625");
+
+	if (firmware_version >= "v0.32") {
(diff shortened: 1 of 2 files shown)


src/analog/m2kanalogin_impl.cpp
```diff
@@ -177,9 +186,9 @@ void M2kAnalogInImpl::setAdcCalibOffset(ANALOG_IN_CHANNEL channel, int raw_offse
 	}
 	const int rawVertOffset = m_adc_hw_offset_raw.at(channel) - m_adc_calib_offset.at(channel);
 	if (m_calibbias_available) {
-                m_adc_calib_offset.at(channel) = m_m2k_adc->setLongValue(channel, raw_offset, "calibbias", false);
```
AlexandraTrifan, line 180: please drop these changes from the commit if no actual changes were done

src/analog/m2kanalogin_impl.hpp
```diff
@@ -180,6 +181,9 @@ class M2kAnalogInImpl : public M2kAnalogIn
 
 	int convertVoltsToRawVerticalOffset(ANALOG_IN_CHANNEL channel, double vertOffset);
 	double convertRawToVoltsVerticalOffset(ANALOG_IN_CHANNEL channel, int rawVertOffset);
+
+	void setCalibrateHDL(std::string calibrate_val);
```
AlexandraTrifan, line 185: could you please provide a short description of this new function, in which fw version was this added and why?

AlexandraTrifan (approved):

PR #353, opened 2023-12-08
tests/validations fw 0.32

Adrian-Stanea:
* Analog test that verifies that no glitch is present with changes made in firmware v0.32
* Investigation test for patter generator known bug that should be addressed in future firmware version..


AlexandraTrifan (approved):
I ran some tests on v0.31 and v0.32 and everything looks good. Can be merged.

PR #382, opened 2025-02-05
matlabBindings: add macos support

IonutMuthi:
Add files needed to create matlab bindings for mac os x86_64 and arm64


bindings/matlab/build_installer.m
```diff
@@ -36,16 +36,26 @@ function build_installer()
 if ispc
     win = 'true';
     unix = 'false';
-else
+    mac = 'false';
+
+elseif isunix
```
Adrian-Stanea, line 41: The current code will return true for both Linux and Mac, therefore the final `else` clause setting the `mac` flag will never be reached.

bindings/matlab/build_installer.m
```diff
@@ -36,16 +36,26 @@ function build_installer()
...
     win = 'false';
     unix = 'true';
+    mac = 'false';
+
+else 
+    win = 'false';
+    unix = 'false';
+    mac = 'true';
 end
+f = strrep(f,'__PROJECT_ROOT__',p);
```
Adrian-Stanea, line 51: This string replacement also requires a change in the bsp.tmpl file because the placeholder we are trying to replace has a new name.

bindings/matlab/build_installer.m
```diff
@@ -36,16 +36,26 @@ function build_installer()
...
 f = strrep(f,'__REPO-ROOT__',p);
 f = strrep(f,'__VERSION__',version);
 f = strrep(f,'__ML-RELEASE__',ml);
 f = strrep(f,'__ARCH__',arch);
 f = strrep(f,'__LINUX__',unix);
 f = strrep(f,'__WINDOWS__',win);
+f = strrep(f,'__APPLE__',mac);
```
Adrian-Stanea, line 58: Same as above, the bsp.tmpl file needs to include macos in the platform tags.

bindings/matlab/build_library.m
```diff
@@ -18,69 +18,111 @@
...
-% build(pkg);
-
-
-
-
-
-
-
-
+% delete definelibm2k.m
```
Adrian-Stanea, line 82: Check if file exists, then delete to avoid warning

Adrian-Stanea (changes requested):

AlexandraTrifan (approved):

Adrian-Stanea (approved):

PR #386, opened 2025-10-06
macOS: Remove CMAKE_OSX_ARCHITECTURES "x86_64"

JJTech0130:
I've been trying to fix arm64 macOS builds of Scopy, and this was causing issues.

Not quite sure why it was added in the first place, clarification would be welcome.


IonutMuthi:
This is already done in main right now so I will close this PR

PR #393, opened 2026-06-10
Fix v1 port for current libiio APIs

999Marius:
## PR Description

Fix buffer, context, and test issues preventing the v1 port from working against current libiio v1 APIs. Several libiio functions used in the original port (`iio_device_create_buffer`, `iio_buffer_enable/disable`, `iio_stream` helper) have since been removed or reworked. These commits update the code to build and pass the full hardware test suite on ADALM2000.

### Benefits

- libm2k v1 port builds and runs correctly against current libiio v1
- Full test suite passes on hardware, and the examples work as expected
- Attribute lookups are correct and use the direct find functions provided by libiio v1

### Changes

- `buffer_v1.cpp`: replace removed APIs with `iio_buffer_stream` interfaces . RX uses a 4-block ring, TX uses a single block. Replace `iio_device_create_buffer` with `iio_device_get_buffer`.
- `buffer_v1.hpp`: replace `iio_stream *m_stream` with `iio_buffer_stream *m_buf_stream` and add ring buffer state (`m_rx_blocks`, `m_rx_head`, `m_current_rx_block`)
- `context_impl_v1.cpp`: use `iio_device_find_attr` / `iio_channel_find_attr` directly instead of manual enumeration with substring matching. Fix `iioDevBufferHasAttribute` which was unconditionally returning `true`, now actually checks buffer attributes.
- `tests/analog_functions.py`: move `aout.push()` before `ctx.setTimeout()` in `test_timeout` , v1 backend doesn't complete stream_start within the short test timeout. Increase `in_samples` from 4096 to 8192 (15 full periods instead of 7.5) to stabilize the average.

### How it works

1. `initializeBuffer` retrieves the device buffer with `iio_device_get_buffer` (replaces the removed `iio_device_create_buffer`).
2. For RX: opens a stream with `iio_buffer_open`, creates 4 blocks via `iio_buffer_stream_create_block`, enqueues all of them, and starts the stream. Each `getSamples` re-enqueues the previously-read block and dequeues the next one.
3. For TX: opens stream, creates a single block, writes data, enqueues, and starts the stream. Non-cyclic dequeues synchronously before returning; cyclic enqueues with cyclic=true and repeats until destroyed.
4. `destroy()` cancels the stream, dequeues blocks, destroys all blocks, and closes the stream.

These changes only affect the v1 backend files (`*_v1.cpp`). The v0 backend is untouched.

### DIFF
--- a/src/context_impl_v1.cpp
+++ b/src/context_impl_v1.cpp
@@ -76,44 +76,22 @@ void ContextImpl::deinitialize()
 
 bool ContextImpl::iioDevHasAttribute(iio_device* dev, std::string const& attr)
 {
-	unsigned int nb_attr = iio_device_get_attrs_count(dev);
-	for (unsigned int i = 0; i < nb_attr; i++) {
-		const struct iio_attr *attr_i = iio_device_get_attr(dev, i);
-		if (attr_i == NULL) {
-			continue;
-		}
-		const char *attr_name = iio_attr_get_name(attr_i);
-		std::size_t found = std::string(attr_name).find(attr);
-		if (found != std::string::npos) {
-			return true;
-		}
-	}
-	return false;
+	return iio_device_find_attr(dev, attr.c_str()) != NULL;
 }
 
 bool ContextImpl::iioDevBufferHasAttribute(iio_device *dev, const std::string &attr)
 {
-	// TBD buffers need to be created in order to check their attributes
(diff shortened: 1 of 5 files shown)


dNechita:
Looks good!
@Adrian-Stanea and @AlexandraTrifan can this be merged into `libiio-v1` branch?

AlexandraTrifan (approved):

PR #394, opened 2026-06-23
CMake: raise minimum version

gastmaier:
Increase CMake minimum required version to 3.10 to avoid warnings on modern CMake about pre-3.10 legacy behavior, and support to CMake 4.


gastmaier (commented):
Looks good to me! Thanks!
For the br2-external I previously overwrote (-DCMAKE_MIN...) to 3.10 to unblock, I will now revert the commit that added the workaround

IonutMuthi (approved):

PR #397, opened 2026-07-28
ci: replaced ubuntu-latest tag with specific ubuntu versions

IonutMuthi:
Instead of ubuntu-latest (currently ubuntu 24.04) now we cover:
* ubuntu 24.04
* ununtu 26.04


liviutomoiaga (approved):
LGTM

PR #402, opened 2026-09-02
buffer_v1: fix false RX dequeue errors over network contexts

999Marius:
iio_block_dequeue() only returns 0 on success for the local backend. Over the network/serial/USB iiod-client backend, a successful dequeue returns the number of bytes transferred, which is a positive, non-zero value.  

Checking `if (ret)` treated every successful network RX dequeue as an error, tearing down the buffer and throwing on the very first getSamples() call.

Check `ret < 0` instead, matching how iiod_client_dequeue_block() actually reports success/failure.


999Marius:
This was fixed in libiio to return 0!

